PR #4, opened 2023-11-08
API Proof of concept

intarga:


intarga:
@lcoram 

The api is defined in api/src/main.rs, with service functions in api/src/{timeseries, timeslice, latest}. The service functions are what I intended to be used (translated) for the frost backend.

Only note I would add is for the timeseries functions: They are currently designed to fetch 1 timeseries at a time, since that's what this API does, but they can be adjusted to support fetching many for the frost use case. Essentially the header query can take a list of station/element ids and use `IN` instead of `=` to match against them, and then you can loop over the results of this to make queries against the data table.

api/src/timeseries.rs
```diff
@@ -0,0 +1,158 @@
...
+    start_time: DateTime<Utc>,
+    end_time: DateTime<Utc>,
+    time_resolution: String,
+) -> Result<TimeseriesRegular, tokio_postgres::Error> {
+    // TODO: string matching intervals like this is a hack, but currently necessary to avoid
+    // SQL injection. Ideally we could pass an interval type as a query param, which would
+    // also save us the query_string allocation, but no ToSql implementations for intervals
+    // currently exist in tokio_postgres, so we need to implement it ourselves.
+    let interval = match time_resolution.as_str() {
+        "PT1M" => "1 minute",
```
lcoram, line 125: PT10M is also pretty standard right now, so maybe that should be here, but its not necessary for the MVP

intarga: Yeah, in theory we should support arbitrary intervals, which a ToSql implementation will achieve. But I felt it wasn't a priority since: 

- the implementation here would be very Rust-specific, and it's your go version that will actually go into production
- The go code going into production in the short term probably won't support regular timeseries anyway

intarga:
@lcoram 

I should also note that this PR includes schema changes, so you will probably need to re-run fake_data_generator to test it

lcoram:
In order to be able to run the fake data generator I had to change the schema path to include "../"

api/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+async fn main() {
+    let args: Vec<String> = std::env::args().collect();
+
+    if args.len() < 4 {
+        panic!("not enough args passed in, at least host, user, dbname needed, optionally password")
+    }
+
+    let mut connect_string = format!("host={} user={} dbname={}", &args[1], &args[2], &args[3]);
+    if args.len() > 4 {
+        connect_string.push(' ');
```
lcoram, line 130: should be " password=" in the string, for password functionality to work

lcoram:
Tested and got this to work:
http://localhost:3000/timeslices/2023-06-02T12:00:00Z/elements/air_temperature 
with a nice simple json output of the results (we would have to deal with level and sensor at some point, but I know for the test we just made them both just 0s... so not in the POC)

http://localhost:3000/stations/10/elements/air_temperature
is giving the error: query returned an unexpected number of rows

http://localhost:3000/latest
returns an empty array, so I might have to check if that is correct... maybe the problem is UTC vs not. 

Investigating these two issues...

intarga:
> http://localhost:3000/stations/10/elements/air_temperature is giving the error: query returned an unexpected number of rows

Pretty sure I know what this is, and good catch. In fake data generator you only generated station_ids between 1000 and 2000 (this causes a different problem of conflicts, which I resolved with `limit 1`), so when you search for station_id 10 it finds no matching time series and errors because of that. We probably need better error handling there to give a more helpful message

lcoram:
I belive for the first issue that at least on my instance the call for the timeseries info does not work...
test-# SELECT timeseries.id, 
                COALESCE(timeseries.fromtime, '1950-01-01 00:00:00+00'), 
                COALESCE(timeseries.totime, NOW()::timestamptz), 
                filter.lvl, 
                filter.sensor, 
                timeseries.loc 
                FROM timeseries JOIN labels.filter 
                    ON timeseries.id = filter.timeseries 
                WHERE filter.station_id = 10 AND filter.element_id = air_temperature 
                LIMIT 1;
ERROR:  syntax error at or near "SELECT"
LINE 2: SELECT timeseries.id,

intarga:
> http://localhost:3000/latest returns an empty array, so I might have to check if that is correct... maybe the problem is UTC vs not.

That’s correct. The fake data generator only generates past data, so there’s nothing within latest_max_age. You will get data if you pass a suitably old date in for latest_max_age

lcoram:
Ah, yes I just realized I was using 10, which is a ts id and not a station number, so thats why its getting nothing... so its not exactly wrong, just a misleading error?

api/src/timeseries.rs
```diff
@@ -0,0 +1,158 @@
...
+                filter.lvl, \
+                filter.sensor, \
+                timeseries.loc \
+                FROM timeseries JOIN labels.filter \
+                    ON timeseries.id = filter.timeseries \
+                WHERE filter.station_id = $1 AND filter.element_id = $2 \
+                LIMIT 1", // TODO: we should probably do something smarter than LIMIT 1
+            &[&station_id, &element_id],
+        )
+        .await?;
```
lcoram, line 60: Could have better error handling here, but unsure if this is a priority for the POC

intarga: Agree, but I’m on holiday now, so I’ll deal with it when I get back

lcoram:
http://localhost:3000/stations/1000/elements/air_temperature?time_resolution=PT1H
works great, and I get that its regular instead of irregular.

http://localhost:3000/latest?latest_max_age=2023-11-03T01:00:00Z
Also got this to work with latest (although that one isn't as fast, guess its doing a decent amount of processing and ends up getting 13264 items...). Looking at the analyze explain for that query makes me wonder if some of the processing is better left up to the rust code and not the sql itself (possibly the distinct aspect... if its still sorted then it should be possible to only take the first for each timeseries?). I feel we do something like this in the old frost, but would have to check exactly how we split up the processing.  

Guess I have checked that things appear to work as expected!

intarga:
> http://localhost:3000/stations/1000/elements/air_temperature?time_resolution=PT1H works great, and I get that its regular instead of irregular.
> 
> http://localhost:3000/latest?latest_max_age=2023-11-03T01:00:00Z Also got this to work with latest (although that one isn't as fast, guess its doing a decent amount of processing and ends up getting 13264 items...). Looking at the analyze explain for that query makes me wonder if some of the processing is better left up to the rust code and not the sql itself (at first i thought maybe the distinct aspect... but removing that actually makes it slower). In frost v0 I am pretty sure we filter for ensuring there is only 1 value, but would have to check exactly how we split up the processing.
> 
> Guess I have checked that things appear to work as expected!

I doubt we could do better than postgres on the processing of latest, postgres is well optimised C code. The reason it's slow is because of how big the result of the index scan is. If you take the default age of 3H, and assume a third of those timeseries are minutely, that's more than 700k rows, your request was probably more.

Processing in postgres also has several advantages we can't replicate in application code:
- Only the final result set needs to be deserialised and passed over the conn, instead of the whole index scan set, which is ~50x bigger. This also saves us >10MiB of heap allocation in the application.
- It can use the DISTINCT constraint and the index on that column to avoid reading blocks from disk if it can see an earlier block fulfilled the criteria.
- It's flexible, if the database statistics change (or we impose new conditions) such that a different query plan would be better, postgres will automatically shift to that with no interaction at all, let alone new code.
- It's less code

The query is definitely intensive enough that we shouldn't expose it directly to public requests, but it could be useful to populate a cache.

api/src/timeseries.rs
intarga: What did you want to change here? The current behaviour is that if the query fails, the server responds with `500 Internal Server Error` + the error string from tokio_postgres. I see two things that can be improved here:

1) Catch expected errors and return a more helpful message to the client. In this case the only expected errors I can think of are database connection failures, and 0 matching timeseries.

2) Address the TODO above, and handle the case of multiple matching timeseries in a more sensible way.

I think 1 is not a priority, since it's really a UX nicety, and this is just a POC that isn't going into production. 2 however would be good to address, since it should also be relevant to your frost backend. 

I need your input though on the best way to solve 2, since I'm not well versed in the edge cases of mapping filter labels to tsids. My understanding is that this would be caused by a station that has been moved or some such. In that case what is the best thing to do? Sort by start date and pick the newest? return all of the timeseries and stitch them together? How does current frost handle this?

lcoram: Yes, 1 is a UX nicety. But I figured maybe it is quite different to get a 500, vs say a 404 (aka, this does not exist). I guess my question is how easily is this differentiable from the errors the code is getting internally? If its not low hanging fruit, then maybe we just leave this whole discussion for the future. 
For a very quick discussion of 2, current Frost caches the mapping between labels and tsids and therefore if they change the cache will be wrong (and need to be reloaded). This was partially because of how slow ODA was, but also generally Frost has had a list of "available timeseries" and their basic definitions (and even from/to times) in order to only actually send as DB call to the database when there is expected to be a return.

intarga:
@lcoram waiting on your approval to merge this

lcoram (approved):

PR #17, opened 2024-04-11
Ansible scripts

lcoram:


lcoram:
closes https://github.com/metno/lard/issues/9

ansible/create-project-network.yml
intarga, line 1: Do these whole files need to be vaulted? Is it not possible to create a secrets file that they use, so the structure can stay open? @lcoram

lcoram: Since the restructure now only the vars file for that particular role is vaulted

lcoram:
Now there are just 2 steps, provision and configure (arnulf mentioned that it was better to seperate those two parts). Also, a way to do the switchover.

lcoram:
Rebased to fix conflicts with changes in db schema

intarga:
> We have tested that Manuel can run this, probably ready to be merged

Yes, sorry I just forgot to do this

intarga (approved):

PR #19, opened 2024-05-21
Add CI and some tests

Lun4m:
- API
   - [x] Endpoint tests

- Ingestor
   - [x] Test `insert_data`
   - [x] Test `timeseries_is_open`
   - [x] Mock permission tables
   - [x] Unit tests
   - [x] ~Test `filter_and_label_kldata`~
   - [x] Test kldata endpoint

- End-to-end
   - [x] Mock permission tables
   - [x] Add example tests
   - [x] Fix bugs


api/src/timeseries.rs
```diff
@@ -127,6 +126,8 @@ pub async fn get_timeseries_data_regular(
         "P1D" => "1 day",
         _ => "1 minute", // FIXME: this should error instead of falling back to a default
     };
+
+    // TODO: this generates nulls till utc.now if end_time is not specified in the database?
```
intarga, line 130: Yes, that's the intended behaviour. If the timeseries hasn't ended then there should be data there, so if there isn't it returns nulls for the gaps

ingestion/src/kldata.rs
```diff
@@ -286,3 +298,174 @@ pub async fn filter_and_label_kldata(
...
+    #[test_case(
+        "kldata/nationalnr=100/type=504/messageid=25" => Ok((100, 504, 25));
+        "valid header 1"
+    )]
+    #[test_case(
+        "kldata/nationalnr=99993/type=508/messageid=23" => Ok((99993, 508, 23));
+        "valid header 2"
+    )]
+    #[test_case(
+        "kldata/nationalnr=93140/type=501/add" => Err(Error::Parse("unexpected field in kldata header format: add".to_string()));
```
lcoram, line 346: What was the weird case that Ingrid talked about yesterday where we have to delete things? do we handle / test this?

lard_tests/tests/end-to-end.rs
```diff
@@ -0,0 +1,115 @@
...
+        let obsinn_msg = format!(
+            "kldata/nationalnr={}/type=508/messageid=23
+TA
+20230505014000,20.0
+20230505014100,20.1
+20230505014200,20.0
+20230505014300,20.2
+20230505014400,20.2
+20230505014500,20.1
+",
```
lcoram, line 33: can / should this be formated differently? I guess maybe not since we can't have spaces in front of the lines... but then the first line could also have no spaces?

ingestion/src/kldata.rs
intarga: It cannot be handled in this PR as it requires changes in obsinn. Issue #21

lard_tests/tests/end-to-end.rs
Lun4m: That's what a raw Obsinn message looks like, but we can for sure construct it in a better way in the tests

Lun4m:
After speaking with Ingrid yesterday, I think I will refactor most of `prepare_db.rs`, `ingestion.rs`, and `api.rs` into different end-to-end tests so we will have everything in one file, while still keeping the same coverage.

lard_tests/tests/end-to-end.rs
```diff
@@ -0,0 +1,115 @@
+use std::future::Future;
```
lcoram, line 1: Most files seem to use snake case, but here you use - not _, is that because its accepted that "end-to-end" is kinda one word?

ingestion/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        insert_data(data, &mut conn).await?;
+
+        Ok(message_id)
+    }
+    .await;
+
+    match result {
+        Ok(message_id) => Json(KldataResp {
+            message: "".into(),
```
lcoram, line 185: help: consider using: `String::new()`

ingestion/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    // set up postgres connection pool
+    let manager = PostgresConnectionManager::new_from_stringlike(connect_string, NoTls)?;
+    let db_pool = bb8::Pool::builder().build(manager).await?;
+
+    // set up param conversion map
+    // TODO: extract to separate function?
+    let param_conversions = Arc::new(
+        csv::Reader::from_path(param_conversion_path)
+            .unwrap()
```
lcoram, line 212: possible panic that needs to be handled? 
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_panics_doc

lard_tests/src/bin/prepare_db.rs
```diff
@@ -0,0 +1,328 @@
...
+    title: &'a str,
+    ts: Timeseries,
+    meta: Labels<'a>,
+}
+
+async fn create_single_ts(client: &Client, ts: Timeseries) -> Result<i32, Error> {
+    let end_time = ts.end_time();
+
+    // Insert timeseries
+    let id = match ts.deactivated {
```
lcoram, line 56: help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#match_bool
help: consider using an `if`/`else` expression

lard_tests/src/bin/prepare_db.rs
```diff
@@ -0,0 +1,328 @@
...
+async fn main() -> Result<(), Error> {
+    // let _test_type: String = env::args()
+    //     .next()
+    //     .expect("Provide test type for database setup ('api', 'ingestion', 'e2e')");
+
+    let (client, connection) = tokio_postgres::connect(CONNECT_STRING, NoTls).await?;
+
+    tokio::spawn(async move {
+        if let Err(e) = connection.await {
+            eprintln!("connection error: {}", e);
```
lcoram, line 318: eprintln!("connection error: {e}");

lard_tests/tests/end-to-end.rs
lcoram: I meant more just for readability, like if the first line was not indented like the rest of them?

ingestion/src/lib.rs
intarga: I think this is fine, this will only panic if the path is wrong/missing, and it will panic at startup.

Did you get this lint from running clippy locally though? Wondering why it doesn't show in CI. If it really is linting though then we should add a doc or annotation to silence it

Cargo.toml
```diff
@@ -2,8 +2,10 @@
 members = [
   "fake_data_generator",
   "api",
-  "ingestion"
+  "ingestion",
+  "lard_tests",
```
intarga, line 6: I would remove lard from the package name and path, to match consistency with the rest of the workspace. Also maybe "integration_tests" would be a better name anyway?

ingestion/src/lib.rs
lcoram: I ran clippy locally in pedantic mode, and kinda scanned through for stuff that I thought was potentially useful or maybe dangerous. I was wrong about this one being relevant, sorry!

intarga: I'm not sure how helpful that is tbh, the things that are gated into pedantic are there because they aren't considered an actual problem, they're nitpicks even by my standards.

lard_tests/tests/end-to-end.rs
Lun4m: I don't know, I'm not a native English speaker, but "end-to-end" seems to be quite well-established! @intarga, should I use underscores instead? For me, it's fine either way.

intarga: https://dictionary.cambridge.org/dictionary/english/end-to-end

It's in the dictionary, so it's definitely established. I don't know what the established practice is for converting hyphenated words to snake case is though, I personally think it would be best to either drop the hyphens entirely or replace them with underscores. My opinion is not that strongly held though, so I won't press it

integration_tests/tests/end_to_end.rs
```diff
@@ -0,0 +1,430 @@
...
+    tokio::spawn(async move {
+        if let Err(e) = conn.await {
+            eprintln!("{}", e);
+        }
+    });
+
+    tokio::select! {
+        _ = api_server => panic!("API server task terminated first"),
+        _ = ingestor => panic!("Ingestor server task terminated first"),
+        // Clean up database even if test panics, to avoid test poisoning
```
intarga, line 209: This is a nice touch

api/src/latest.rs
```diff
@@ -1,13 +1,13 @@
...
+#[derive(Debug, Serialize, Deserialize)]
 pub struct LatestElem {
-    value: f32,
-    timestamp: DateTime<Utc>,
-    station_id: i32,
-    loc: Option<Location>,
+    pub value: f32,
+    pub timestamp: DateTime<Utc>,
+    pub station_id: i32,
+    pub loc: Option<Location>,
```
intarga, line 10: @Lun4m is there a reason all these struct fields need to be pub? I can't see where you use them

Lun4m: Just convenience for the future, but yes, it makes more sense to make them public when we write new tests that actually use them.

intarga (approved):

PR #29, opened 2024-09-24
Add ingestion of non-scalar parameters

Lun4m:
This PR fixes the obsinn header parsing and implements the ingestion of non-scalar parameters to a separate `nonscalar_data` table, where for now the `obsvalue` has type `TEXT`. Once we have a clearer idea of what types of non-scalars we are dealing with and how they are used downstream we can come up with better database design.

Edit: the number of lines affected has mostly to do with the fact that I added a new column to `paramconversions.csv`.


ingestion/src/kldata.rs
```diff
@@ -35,47 +44,84 @@ struct ObsinnId {
...
-        ))
-    })?;
+// TODO: maybe this can be a field in ObsinnChunk?
+#[derive(Default)]
+struct ObsinnHeader {
+    station_id: Option<i32>,
+    type_id: Option<i32>,
+    message_id: usize,
+    // TODO: we can parse it to Datatime if we decide we are going to use it
+    received_time: Option<String>,
```
intarga, line 54: I don't think these fields should be `Option`, since the only place they can be `None` is inside the constructor. I think it would be better to remove the `Option` here and use separate variables (which themselves are `Option`) in the contructor, and build the struct like:

```
return ObsinnHeader {
    station_id: station_id.ok_or(...)?,
    ...
```

ingestion/src/kldata.rs
```diff
@@ -35,47 +44,84 @@ struct ObsinnId {
...
+    }
+}
+
+fn parse_value<T: FromStr>(key: &str, value: &str) -> Result<T, Error>
+where
+    <T as FromStr>::Err: Debug,
+{
+    value
+        .parse::<T>()
+        .map_err(|_| Error::Parse(format!("invalid number in kldata header for key {}", key)))
```
intarga, line 124: Nitpick, but I think "number" here is now technically wrong, since you're using this to parse strings now for received_time

ingestion/src/kldata.rs
intarga, line 54: This can maybe be &str instead of String? since it's just referencing a substring of the message body?

ingestion/src/kldata.rs
```diff
@@ -35,47 +44,84 @@ struct ObsinnId {
...
-    let next_err = || Error::Parse("kldata header terminated early".to_string());
+        let mut header = ObsinnHeader::default();
 
-    if parts.next().ok_or_else(next_err)? != "kldata" {
-        return Err(Error::Parse(
-            "kldata indicator missing or out of order".to_string(),
-        ));
-    }
+        for field in fields.by_ref() {
+            // TODO: this field signals data deletion/update in kvalobs, we do not use it
```
intarga, line 81: ??? Where did you find out about this from? How does it work? I thought our conversations with Søren concluded that messages are implicitly deletions/updates if that (station_id, timestamp) had been sent before?

ingestion/src/kldata.rs
```diff
@@ -106,66 +152,84 @@ fn parse_columns(cols_raw: &str) -> Result<Vec<ObsinnId>, Error> {
         .collect::<Result<Vec<ObsinnId>, Error>>()
 }
 
-fn parse_obs(csv_body: Lines<'_>, columns: &[ObsinnId]) -> Result<Vec<ObsinnObs>, Error> {
+fn parse_obs(
+    csv_body: Lines<'_>,
+    columns: &[ObsinnId],
+    params: Arc<HashMap<String, Param>>,
```
intarga, line 158: I think we should give this argument a better name to make clear that it doesn't refer to the params in the message (which are in fact encoded in `columns`) but rather our general reference list of params. Perhaps "params_reference" or "param_types"?

ingestion/src/kldata.rs
```diff
@@ -106,66 +152,84 @@ fn parse_columns(cols_raw: &str) -> Result<Vec<ObsinnId>, Error> {
...
+        let mut vals = row.split(',');
 
-            // TODO: timestamp parsing needs to handle milliseconds and truncated timestamps?
-            let timestamp = NaiveDateTime::parse_from_str(raw_timestamp, "%Y%m%d%H%M%S")
-                .map_err(|e| Error::Parse(e.to_string()))?
-                .and_utc();
+        let raw_timestamp = vals.next().ok_or_else(row_is_empty)?;
 
-            (timestamp, vals)
-        };
```
intarga, line 125: Any particular reason you removed the scoping here? The original reason for the scoping was to limit where `vals` is mut to only the timestamp extraction

ingestion/src/kldata.rs
```diff
@@ -106,66 +152,84 @@ fn parse_columns(cols_raw: &str) -> Result<Vec<ObsinnId>, Error> {
...
+            .and_utc();
 
         for (i, val) in vals.enumerate() {
-            // Should we do some smart bounds-checking??
+            // TODO: should we do some smart bounds-checking??
             let col = columns[i].clone();
 
-            // TODO: parse differently based on param_code?
+            let value = {
+                let param = params.get(&col.param_code).ok_or_else(|| {
```
intarga, line 178: this name should similarly be changed if we change `params`

ingestion/src/kldata.rs
```diff
@@ -106,66 +152,84 @@ fn parse_columns(cols_raw: &str) -> Result<Vec<ObsinnId>, Error> {
...
+                    Error::Parse(format!("unrecognised param_code '{}'", col.param_code))
+                })?;
```
intarga, line 180: There's maybe a case to be made for just logging this error instead of returning, and treating unknown params like the scalar params (which we're already treating like we don't know their type)

ingestion/src/kldata.rs
```diff
@@ -106,66 +152,84 @@ fn parse_columns(cols_raw: &str) -> Result<Vec<ObsinnId>, Error> {
...
+
+                if param.is_scalar {
```
intarga, line 182: Maybe worth leaving a note here that we're assuming assuming params marked by stinfosys as "scalar" can be parsed as float, but some may turn out to be integers

ingestion/src/kldata.rs
```diff
@@ -106,66 +152,84 @@ fn parse_columns(cols_raw: &str) -> Result<Vec<ObsinnId>, Error> {
...
+                    })?;
+
+                    ObsType::Scalar(parsed)
+                } else {
+                    // TODO: we should implement logging/tracing sooner or later
+                    println!(
+                        "Non-scalar param ({}, {}, {}): '{}'",
+                        param.id, col.param_code, param.element_id, val
+                    );
+                    ObsType::NonScalar(val.to_string())
```
intarga, line 194: I think the type here can be &str, it would save us having to do an allocation for every non-scalar datum

ingestion/src/kldata.rs
```diff
@@ -106,66 +152,84 @@ fn parse_columns(cols_raw: &str) -> Result<Vec<ObsinnId>, Error> {
...
-        let columns = parse_columns(lines.next().ok_or_else(lines_err)?)?;
+pub fn parse_kldata(
+    msg: &str,
+    param_conversions: Arc<HashMap<String, Param>>,
+) -> Result<(usize, ObsinnChunk), Error> {
+    let mut csv_body = msg.lines();
+    let lines_err = || Error::Parse("kldata message contained too few lines".to_string());
 
-        (station_id, type_id, message_id, columns, lines)
-    };
```
intarga, line 161: Here again I don't think we should remove the scoping

ingestion/src/lib.rs
```diff
@@ -53,7 +53,14 @@ pub type PgConnectionPool = bb8::Pool<PostgresConnectionManager<NoTls>>;
 
 pub type PooledPgConn<'a> = PooledConnection<'a, PostgresConnectionManager<NoTls>>;
 
-type ParamConversions = Arc<HashMap<String, (String, i32)>>;
+#[derive(Clone, Debug)]
+pub struct Param {
```
intarga, line 57: I already raised this when this was used as a param, but I think this could be named more clearly

ingestion/src/lib.rs
```diff
@@ -53,7 +53,14 @@ pub type PgConnectionPool = bb8::Pool<PostgresConnectionManager<NoTls>>;
...
+    id: i32,
+    element_id: String,
+    is_scalar: bool,
```
intarga, line 60: Perhaps it would be nice to have a brief comment providing context for each of these?

ingestion/src/lib.rs
```diff
@@ -130,7 +158,7 @@ pub async fn insert_data(data: Data, conn: &mut PooledPgConn<'_>) -> Result<(),
 }
 
 pub mod kldata;
-use kldata::{filter_and_label_kldata, parse_kldata};
+use kldata::{filter_and_label_kldata, parse_kldata, ObsType};
```
intarga, line 161: I think it would make sense for ObsType to live outside the kldata mod (I guess we will need to create a util mod so kldata can import it), since it's general to ingestion, not specific to kldata/obsinn, and other future sources will need it too.

ingestion/src/lib.rs
```diff
@@ -84,11 +91,12 @@ impl FromRef<IngestorState> for Arc<RwLock<(ParamPermitTable, StationPermitTable
 pub struct Datum {
     timeseries_id: i32,
     timestamp: DateTime<Utc>,
-    value: f32,
+    value: ObsType,
 }
 
 pub type Data = Vec<Datum>;
 
+// TODO: benchmark insertion of scalar and non-scalar together vs separately?
```
intarga, line 99: Yes, that would be a good idea. I suspect postgres is smart enough to handle that on its own, but it would be good to know for sure. In any case, it would be easy to put scalar and nonscalar params in different vecs, and it would save us the branching here

integration_tests/src/main.rs
```diff
@@ -9,13 +9,31 @@ async fn insert_schema(client: &tokio_postgres::Client, filename: &str) -> Resul
     client.batch_execute(schema.as_str()).await
 }
 
-async fn create_data_partition(client: &tokio_postgres::Client) -> Result<(), Error> {
-    // TODO: add multiple partitions?
```
intarga, line 13: Idk if it's intentional that you dropped this todo

intarga (changes requested):
Nice work!

ingestion/src/kldata.rs
Lun4m: Yep, I thought about that but then forgot to change them. I'm only keeping `received_time` as on `Option` since the message is still valid if that field is missing

Lun4m: True, I just didn't want to add explicit lifetimes

ingestion/src/kldata.rs
Lun4m: From a conversation with Søren that I should have probably recorded :rofl: I seem to recall that when the `add` keyword is specified the values for the different timestamps are updated only for the given `param_codes` and deleted for the others, but that doesn't sound quite right, I'll try to ask him again :facepalm:

integration_tests/src/main.rs
Lun4m: I guess there isn't really a point in having multiple partitions here unless we want to perform specific tests that make use of them. But yes, I think it's probably worth keeping the TODO

ingestion/src/lib.rs
intarga: Yeah that's good 👍

intarga (approved):

ingestion/src/kldata.rs
```diff
@@ -23,9 +23,18 @@ pub struct ObsinnChunk {
 /// Represents a single observation from an obsinn message
 #[derive(Debug, PartialEq)]
 pub struct ObsinnObs {
+    // TODO: this timestamp is shared by all obs in the row, maybe we should have
+    // a Vec<ObsType> here and remove the Vec<ObsinnObs> from ObsinnChunk
```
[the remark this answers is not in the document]
intarga: To be clear, I'm happy to leave this as a todo, though it wouldn't hurt to add some context from this discussion

ingestion/src/lib.rs
intarga: Also happy to leave this as a todo

ingestion/src/kldata.rs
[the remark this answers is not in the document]
Lun4m: Ok, will do and then merge!

PR #42, opened 2024-12-11
Use rove for QC in the ingestor

intarga:
- [x] Schema changes to support flags
- [x] Plumbing
- [x] `qc_data` implementation
- [x] identify time_resolution from type_id
- [x] load and use real pipeline
- [x] Make sure mock pipeline is consistent with the real one
- [x] Merge confident.flags into public.data

This PR integrates [rove](https://github.com/metno/rove) to perform QC in the ingestor with the current limitations:

- Only fresh pipelines are supported (not periodic or consistency)
- We only have 1 actual pipeline (hourly air temperature), and even that has 2 checks disabled as we don't have access to model data
- No caching of prior ingested observations

I considered verifying the QC output in the integration tests, but decided this was best left for when we implement an API that actually uses them (frost backend?) for now I manually queried the DB after running the integration tests to confirm `qc_usable` was as expected.

I restructured the generic `Datum` and `DataChunk` to match their obsinn equivalents. The original idea behind the different structure was that I thought they should be neutral towards different data sources that might structure messages differently, but I realised that the Obsinn structure of clustering by station and timestamp is what makes sense for any source of fresh data, so we should optimise for that.

There is [a smaller PR of associated fixes in ROVE](https://github.com/metno/rove/pull/94) that should be merged before this one


Cargo.toml
```diff
@@ -27,8 +27,10 @@ quick-xml = { version = "0.35.0", features = [ "serialize", "overlapped-lists" ]
 rand = "0.8.5"
 rand_distr = "0.4.3"
 regex = "1.11.1"
-rove = { git = "https://github.com/metno/rove.git" }
+rove = { git = "https://github.com/metno/rove.git", branch = "lard_fixes" }
+# rove = { git = "https://github.com/metno/rove.git" }
```
intarga, line 31: This needs to be removed before merging

ingestion/src/lib.rs
```diff
@@ -170,6 +269,67 @@ pub async fn insert_data(data: Data<'_>, conn: &mut PooledPgConn<'_>) -> Result<
...
+            let rove_output = rove::Scheduler::schedule_tests(pipeline, data)?;
+
+            let first_fail = rove_output.iter().find(|check_result| {
+                if let Some(result) = check_result.results.first() {
+                    if let Some(flag) = result.values.first() {
+                        return *flag == rove::Flag::Fail;
+                    }
+                }
+                false
+            });
```
Lun4m, line 310: Here you are using `first` because we expect a single timeseries and single data point? I would either add a comment about it or generalize it. And do we plan to use this same function to run periodic/consistency checks? If not. I'd simply rename this one to `qc_fresh_data`

ingestion/src/lib.rs
```diff
@@ -170,6 +269,67 @@ pub async fn insert_data(data: Data<'_>, conn: &mut PooledPgConn<'_>) -> Result<
...
+
+            let (flag, fail_condition) = match first_fail {
+                Some(check_result) => (1, Some(check_result.check.clone())),
+                None => (0, None),
+            };
+
+            datum.qc_usable = Some(flag == 0);
```
Lun4m, line 317: `DataMissing` and `Warn` are not valid flags here?

integration_tests/tests/end_to_end.rs
```diff
@@ -245,6 +246,11 @@ async fn e2e_test_wrapper<T: Future<Output = ()>>(test: T) {
     let api_pool = db_pool.clone();
     let ingestion_pool = db_pool.clone();
 
+    let rove_connector = Connector {
+        pool: db_pool.clone(),
+    };
+    let qc_pipelines = load_pipelines("mock_qc_pipelines/fresh").expect("failed to load pipelines");
```
Lun4m, line 252: Why do we need a separate file here? Can't we simply use `qc_pipelines/fresh/TA_PT1H.toml`? We probably need a helper method to load a single file though.

util/src/bin/generate_partition_queries.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+fn main() -> Result<(), std::io::Error> {
+    let outfile = File::create("../db/partitions_generated.sql")?;
+    let mut writer = BufWriter::new(outfile);
+
+    writer.write_all("-- Generated by util/src/bin/generate_partition_queries.rs\n".as_bytes())?;
+
+    // create a vector of the boundaries between partitions
+    let paritition_boundary_years: Vec<DateTime<Utc>> = [1850, 1950, 2000, 2010]
```
Lun4m, line 40:
```suggestion
    let paritition_boundary_years: Vec<DateTime<Utc>> = [1700, 1850, 1950, 2000, 2010]
```

Some migration data is apparently older than 1850 :shaking_face:

integration_tests/src/main.rs
```diff
@@ -24,9 +24,9 @@ async fn main() {
     // NOTE: order matters
```
Lun4m, line 24: Should we move this file to the `util` crate?

db/flags.sql
```diff
@@ -1,5 +1,20 @@
...
+CREATE TABLE IF NOT EXISTS flags.confident_provenance (
+    timeseries INT4 NOT NULL,
+    obstime TIMESTAMPTZ NOT NULL,
+    pipeline TEXT NOT NULL,
+    -- TODO: should this be an enum?
+    flag INT4 NOT NULL,
+    -- TODO: better name? since this might be applied to flags that aren't fail but also aren't pass?
+    fail_condition TEXT NULL,
+    CONSTRAINT unique_confident_providence_timeseries_obstime_pipeline UNIQUE (timeseries, obstime, pipeline),
+    CONSTRAINT fk_confident_providence_timeseries FOREIGN KEY (timeseries) REFERENCES public.timeseries
```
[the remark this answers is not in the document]
intarga: Damn it, I thought I fixed those typos! I guess not

ingestion/src/kldata.rs
```diff
@@ -262,116 +276,126 @@ pub async fn filter_and_label_kldata<'a>(
...
+                .ok_or_else(|| {
+                    Error::Parse(format!(
+                        "unrecognised param_code '{}'",
+                        in_datum.id.param_code
+                    ))
+                })?;
+
+            // TODO: we only need to check inside this loop if station_id is in the
+            // param_permit_table
+            if !timeseries_is_open(
```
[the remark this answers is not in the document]
intarga: Do you want me to expand on the comment?

ingestion/src/lib.rs
```diff
@@ -125,41 +179,86 @@ pub async fn insert_data(data: Data<'_>, conn: &mut PooledPgConn<'_>) -> Result<
...
     let query_nonscalar = conn
         .prepare(
-            "INSERT INTO public.nonscalar_data (timeseries, obstime, obsvalue) \
-                VALUES ($1, $2, $3) \
+            "INSERT INTO public.nonscalar_data (timeseries, obstime, obsvalue, qc_usable) \
+                VALUES ($1, $2, $3, $4) \
                 ON CONFLICT ON CONSTRAINT unique_nonscalar_data_timeseries_obstime \
-                    DO UPDATE SET obsvalue = EXCLUDED.obsvalue",
+                    DO UPDATE SET obsvalue = EXCLUDED.obsvalue, \
+                    qc_usable = public.nonscalar_data.qc_usable AND EXCLUDED.qc_usable",
```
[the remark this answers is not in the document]
intarga: Are we sure of this? I know some consistency checks cover cloud conditions which aren't floats (though they may be encoded as floats in kvalobs, I'm not sure). I feel somewhat inclined to keep this anyway, because it's feasible that future checks or HQC might look at them, right?

ingestion/src/lib.rs
```diff
@@ -170,6 +269,67 @@ pub async fn insert_data(data: Data<'_>, conn: &mut PooledPgConn<'_>) -> Result<
...
+    chunks: &mut Vec<DataChunk<'_>>,
+    rove_connector: &rove_connector::Connector,
+    pipelines: &HashMap<(i32, RelativeDuration), rove::Pipeline>,
+) -> Result<Vec<QcProvenance>, Error> {
+    let mut qc_results: Vec<QcProvenance> = Vec::new();
+    for chunk in chunks {
+        let time_resolution = match chunk.time_resolution {
+            Some(time_resolution) => time_resolution,
+            // if there's no time_resolution, we can't QC
+            None => continue,
```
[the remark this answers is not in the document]
intarga: At the moment the pipelines are keyed by (param_id, time_resolution), so it definitely doesn't make sense to keep going without a time resolution.

I guess in theory we could specify time_resolution for a pipeline as `"*"` in the header, but:
- Line has not yet offered any clarity on what the header spec needs to cover
- I'm not convinced it's really that useful because single values of different time resolutions aren't always (or even usually) equivalent (e.g. precipitation)
- How likely are we to have a pipeline that only consists of single checks?

I'd rather wait until this becomes necessary to implement by the 3 conditions above being satisfied by a pipeline

ingestion/src/lib.rs
```diff
@@ -170,6 +269,67 @@ pub async fn insert_data(data: Data<'_>, conn: &mut PooledPgConn<'_>) -> Result<
...
+                None => continue,
+            };
+            let data = rove_connector
+                .fetch_one(
+                    datum.timeseries_id,
+                    &time_spec,
+                    pipeline.num_leading_required,
+                    pipeline.num_trailing_required,
+                )
+                .await?;
```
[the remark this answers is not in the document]
intarga: > but we are not QCing the current piece of data, datum.value (which is not yet in lard at this point)?

Ahhhh good catch! In the original implementation I had flags in a separate table, so the data was inserted by this point.

I guess it would be possible to mutate the result of `fetch_one` to put `datum.value` into it. Actually now that I'm thinking about it it doesn't make sense to have trailing points on a fresh pipeline, so we should probably disallow those, explicitly fetch only the leading points, and append the current value.

ingestion/src/lib.rs
intarga: > Here you are using first because we expect a single timeseries and single data point?

Yes

> And do we plan to use this same function to run periodic/consistency checks?

No, I think their ideal signatures are too different

I'll add clarifying comments, and rename the function

integration_tests/tests/end_to_end.rs
intarga: My thought behind this was that in the real pipelines, params/steps might change such that the QC output changes, and we eventually want to be checking QC output in the integration tests. That could result in param/step changes breaking CI in ways they shouldn't.

While this isn't true right now (the mock and real pipelines are currently identical), it will be true in the future. Additionally, people updating pipeline definitions will eventually use CI to verify their pipelines are valid. I don't want them to be thrown off by their changes causing unrelated CI failures.

integration_tests/src/main.rs
intarga: Hmm... Because you are thinking of using it from ansible? I would move it to the util crate once it's actually used outside this one, rather than before

ingestion/src/lib.rs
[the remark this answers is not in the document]
Lun4m: Hmm, I know for sure that the `text_data` table in Kvalobs doesn't have the `corrected`, `controlinfo`, `useinfo` and `cfailed` columns. Maybe they use the cloud types as "extra" data for the checks but they don't actually QC them?

ingestion/src/lib.rs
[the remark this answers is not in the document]
Lun4m: Fair enough, my doubt was more about how to QC data coming in with a typeid that doesn't match a specific time resolution. But I agree, we would need to see a pipeline for those param_id, type_id combos to decide what's best to do

integration_tests/tests/end_to_end.rs
Lun4m: I'm a bit confused by the last two sentences, but how can they verify that the changes are valid if they are not tested? Maybe I'm just sleepy :sweat_smile:

ingestion/src/kldata.rs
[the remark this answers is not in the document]
Lun4m: That would be nice :)

ingestion/src/lib.rs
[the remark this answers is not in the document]
intarga: The plan is eventually to have [more] complete and correct time resolution info in the timeseries or met label table, which would let us grab that during the labelling, so we'd then have it here (except perhaps for very recently commissioned stations, which can be covered retroactively by periodic checks). And extra eventually perhaps even get the new time resolution format directly from obsinn / the station network.

integration_tests/tests/end_to_end.rs
intarga: Sorry, I was unclear. In the future, there will be a new test that loads all the real pipelines, checks they deserialise correctly, and perhaps also provides some linting

db/flags.sql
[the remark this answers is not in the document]
intarga: Done

ingestion/src/lib.rs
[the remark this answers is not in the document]
intarga: Done. I made a new method on the connector, we should probably use that to inform the future evolution of the `DataConnector` trait.

ingestion/src/kldata.rs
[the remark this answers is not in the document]
intarga: Expanded, let me know if the new comment makes sense.

db/public.sql
```diff
@@ -32,6 +32,9 @@ CREATE TABLE IF NOT EXISTS public.data (
     timeseries INT4 NOT NULL,
     obstime TIMESTAMPTZ NOT NULL,
     obsvalue REAL,
+    -- TODO: should qc_usable be NOT NULL? and default to true?
+    -- It would make greatly reduce the update load when QCing old data
+    qc_usable BOOLEAN,
```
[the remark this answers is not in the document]
intarga: Hmm, I'm inclined to think we should try to address it now, as I'm not sure we will get any more clarity on the uncertainties.

From my perspective, this comes down to whether a `NULL` here would actually be useful to anyone. I think Frost or any other API would have to treat `NULL` and `true` identically, since many params aren't QCed at all (and these are params should be the only `NULL`s anyway). Edge case users who want to know specifically whether a piece of data has been QCed will likely want to look at the provenance table anyway, since they probably want to also know which pipelines have been run; a lack of entries in that table would indicate the data has not been QCed.

For these reasons I lean towards `NOT NULL`. My only caveat is that it should probably be documented in the API (when one that uses this eventually exists) that `qc_usable = true` only implies that an observation didn't fail QC, not that QC has been run on it.

ingestion/src/lib.rs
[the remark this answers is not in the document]
intarga: Hmm, that seems pretty certain they don't have this in kvalobs. Unless you feel strongly about removing it though, I feel inclined to keep it for the reasons stated above, plus keeping the data tables as structurally similar as possible.

ingestion/src/lib.rs
intarga: Done

util/src/bin/generate_partition_queries.rs
intarga: Done

Cargo.toml
intarga: Done

db/public.sql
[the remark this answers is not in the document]
Lun4m: The only problem I see is that this way you can't really trust a `true` value. The user fetches all data with `true`, then some QC is run and some end up `false`. Now the user is stuck with invalid data, but they don't know it. It sounds a bit dangerous :sweat_smile: If we know which params are not to be QCed, it should be easy to implement a different query for them in Frost.

But now that I think about it Frost needs to compute the `observation_quality` id (or whatever it's called), so it might look into the provenance table by default anyway? Or is the `qc_usable` column meant to replace that? Maybe it should be an enum then?

Sorry I'm jumping all over the place, if I sound confusing we can have a zoom chat sometime :)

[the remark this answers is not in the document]
intarga: > The only problem I see is that this way you can't really trust a true value. The user fetches all data with true, then some QC is run and some end up false.

I think this is already the case though. If a user fetches recent data that's only had fresh QC run on it, some of the data marked with `qc_usable = true` might be later overwritten to false by a periodic pipeline. Even for old data, it can be re-QCed with new check parameters. I think we just have to accept that what we report from the API can only reflect our current knowledge of the data, not absolute truth about its quality.

> But now that I think about it Frost needs to compute the observation_quality id (or whatever it's called), so it might look into the provenance table by default anyway? Or is the qc_usable column meant to replace that? Maybe it should be an enum then?

For now Frost will use the kvalobs QC to derive `qualityIndex` as it currently does. When we switch to using confident QC, this bool[s] will replace `qualityIndex` (though we may decide to still offer a `qualityIndex` option in the API for backwards compatibility, this would likely only ever be 1 of 2 values that would map directly from the bool).

> we can have a zoom chat sometime

Your comment was actually very clear. I'm going to try to come to the office tomorrow though

ingestion/src/kldata.rs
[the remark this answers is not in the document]
Lun4m: Looks good! :ok_hand:

ingestion/src/lib.rs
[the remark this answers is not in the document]
Lun4m: Alright!

db/public.sql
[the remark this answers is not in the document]
Lun4m: > I think this is already the case though. If a user fetches recent data that's only had fresh QC run on it, some of the data marked with `qc_usable = true` might be later overwritten to false by a periodic pipeline. Even for old data, it can be re-QCed with new pipeline configuration. I think we just have to accept that what we report from the API can only reflect our current knowledge of the data, not absolute truth about its quality.

Ah, that's a good point, and while it's not exactly the same, it doesn't sound that dangerous anymore.

Regarding the quality index, weren't we supposed to have multiple usability conditions? Or was that idea dropped?

Lun4m (approved):

db/public.sql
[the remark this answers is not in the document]
intarga: > Regarding the quality index, weren't we supposed to have multiple usability conditions? Or was that idea dropped?

Yes, but it requires usecase-specific flag conditions to be implemented in the pipelines; we're waiting on Line for that. For now we're just treating any check failure as a pipeline failure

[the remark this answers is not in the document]
intarga: I've updated this to default to true, and added some comments to explain the rationale.

I had to modify the migrations to comply with this change, I guess we can discuss the implications of that tomorrow. 

I was surprised to [re-?]discover that `go build` doesn't prompt you for missing struct fields, it just fills them in with the default value for their type. Jesus Christ that's a sharp edge...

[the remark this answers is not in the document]
Lun4m: Yeah, that's not great :disappointed: But If you added default true on the table definition you wouldn't have had to touch the migration code, is there any benefit/disadvantage of doing it this way? And now that I think about it, I don't have to modify the migration code to calculate `qc_usable` from the kvalobs flags right??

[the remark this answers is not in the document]
intarga: > But If you'd added default true on the table definition you wouldn't have had to touch the migration code

Good point, I'll do that and drop the migration changes

> I don't have to modify the migration code to calculate qc_usable from the kvalobs flags right

No, qc_usable is only for confident QC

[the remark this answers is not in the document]
intarga: Done

db/flags.sql
```diff
@@ -1,5 +1,20 @@
 CREATE SCHEMA IF NOT EXISTS flags;
 
+-- TODO: should this also have a column for qc_time or some such?
+CREATE TABLE IF NOT EXISTS flags.confident_provenance (
+    timeseries INT4 NOT NULL,
+    obstime TIMESTAMPTZ NOT NULL,
+    pipeline TEXT NOT NULL,
+    -- TODO: should this be an enum?
```
jo-asplin-met-no, line 8: Does this comment refer to the line below or the one below? Suggest to move the comment to after the line to remove ambiguity.

db/partitions_generated.sql
```diff
@@ -1,77 +1,115 @@
--- Generated by simple script for testing
-CREATE TABLE IF NOT EXISTS data_y1850_to_y1950 PARTITION OF public.data
-FOR VALUES FROM ('1850-01-01 00:00:00+00') TO ('1950-01-01 00:00:00+00');
-CREATE TABLE IF NOT EXISTS data_y1950_to_y2000 PARTITION OF public.data
+-- Generated by util/src/bin/generate_partition_queries.rs
```
jo-asplin-met-no, line 1: What's the purpose of all these tables? And why is the time dimension partitioned exactly like that?

db/flags.sql
```diff
@@ -1,5 +1,20 @@
 CREATE SCHEMA IF NOT EXISTS flags;
 
+-- TODO: should this also have a column for qc_time or some such?
+CREATE TABLE IF NOT EXISTS flags.confident_provenance (
```
jo-asplin-met-no, line 4: Does this mean that LARD now supports provenance? And if so at what level? QC update history of individual observations?

db/public.sql
```diff
@@ -32,6 +32,12 @@ CREATE TABLE IF NOT EXISTS public.data (
     timeseries INT4 NOT NULL,
     obstime TIMESTAMPTZ NOT NULL,
     obsvalue REAL,
+    -- This value should be treated as an absolute assertion of the data's quality but rather our
```
jo-asplin-met-no, line 35: Guess it should say: "... should _not_ be treated ..."

ingestion/src/kldata.rs
```diff
@@ -262,116 +276,131 @@ pub async fn filter_and_label_kldata<'a>(
...
+                continue;
+            }
 
-                // create obsinn label
-                transaction
-                    .execute(
-                        "INSERT INTO labels.obsinn \
+            let transaction = conn.transaction().await?;
+
+            let (sensor, lvl) = in_datum
```
jo-asplin-met-no, line 313: Any particular reason for using _lvl_ instead of _level_? The latter is more readable.

db/partitions_generated.sql
intarga: The data table is very large (>1 billion rows), and such large tables have performance problems because their indices grow too large and become expensive to use and especially to rebalance. Partitioning is an attempt to address that, as the partitions have separate indices.

`obstime` seems like an obvious choice for a partition key since
- It can be partitioned predictably (it's easy to get your partitions to be a certain size and control their growth)
- It offers good query locality on a single partition which will keep its indices hot
- It's already btree indexed

As for the partition boundaries, they're non linear because the goal is to have each partition contain approximately the same number of rows, and time-density of our dataset has increased over time. The specific numbers are just guesses for now; @Lun4m is planning to measure and rebalance the boundaries when he's migrated all of KDVH/Kvalobs

db/flags.sql
intarga: > Does this mean that LARD now supports provenance?

For confident QC only, yes. As it's confident only this will not be relevant to the beta, which will be looking at the flags.kvdata table for both data and QC (I know it's confusing that it's in the flags schema, we've discussed this and are probably going to have it to the public schema).

> And if so at what level? QC update history of individual observations?

As the table definition encodes: For each piece of data that's QCed, it will have a row here for each type of pipeline run on it (fresh, periodic, consistency). This is a compromise between trying to keep the information that's actually useful, without having this table be unusably large. Worth noting that this table design is prospective, subject to change upon feedback from the confident project.

db/public.sql
intarga: Good catch, fixed!

ingestion/src/kldata.rs
intarga: I believe we used `lvl` to match what it's called in obsinn/kvalobs. No objection from me if you feel strongly about changing it, I agree `level` is more readable.

db/partitions_generated.sql
jo-asplin-met-no: Will the boundary rebalancing be straightforward once the system gets truly operational? And will it require much downtime of the system etc.?

ingestion/src/kldata.rs
```diff
@@ -234,22 +236,34 @@ pub fn parse_kldata(
...
-            type_id: header.type_id,
-        },
+        parse_obs(csv_body, &columns, reference_params, header)?,
     ))
 }
 
+// TODO: this is a messy hack, but it's the only way people at met currently have to determine
+// time_resolution. Ultimately we intend to store time_resolution info in the database under
+// public.timeseries or labels.met. This will be populated by a combination of a script that looks
+// at a timeseries's history, and manual editing by content managers.
```
jo-asplin-met-no, line 246: Q1: Does this mean that at that point time resolution info is no longer maintained in stinfosys? (not too familiar with stinfosys, but I believe the type_id->time_resolution mapping is kept there somehow).
Q2: Will lard support time series without a fixed time resolution? (I believe (Cassandra-based) ODA does, but I once heard an performance-related argument against it; I do have the feeling that a mandatory time resolution could be too restrictive for some use cases, but then a compromise could perhaps be keep that kind of data in a separate (and slower!) part of the database. But as always, if there really aren't any use cases for it, we probably shouldn't bother.)

db/partitions_generated.sql
Lun4m: Once it's operational, we will only have to estimate (in advance) the boundaries of new partitions, so there shouldn't be any downtime. It could be also be done by a cron job, but we haven't looked it into yet.

ingestion/src/kldata.rs
```diff
@@ -18,17 +19,12 @@ pub struct ObsinnChunk<'a> {
     observations: Vec<ObsinnObs<'a>>,
```
jo-asplin-met-no, line 19: General question about the term 'obsinn'. Does this mean that the dependency between lard and Obsinn goes both ways, i.e. that also lard needs to know about obsinn-specific formats and names? While this is probably fine, I'm just wondering if the design will also support other input sources than Obsinn. Will those input sources then have to mimic Obsinn somehow to be ingestible by lard?

ingestion/src/kldata.rs
jo-asplin-met-no: No it's fine, it's not that unreadable I guess. And it's a good point that it should match obsinn/kvalobs. But see my other comment about dependency to obsinn in general.

ingestion/src/kldata.rs
```diff
@@ -262,116 +276,131 @@ pub async fn filter_and_label_kldata<'a>(
...
+                    ],
+                )
+                .await?;
+
+            let timeseries_id: i32 = match obsinn_label_result {
+                Some(row) => row.get(0),
+                None => {
+                    // create new timeseries
+                    // TODO: currently we create a timeseries with null location
+                    // In the future the location column should be moved to the timeseries metadata table
```
jo-asplin-met-no, line 337: Will/does lard also support per-obs location? (required for moving sensors)

ingestion/src/lib.rs
```diff
@@ -73,6 +79,8 @@ struct IngestorState {
     db_pool: PgConnectionPool,
     param_conversions: ParamConversions, // converts param codes to element ids
     permit_tables: Arc<RwLock<(ParamPermitTable, StationPermitTable)>>,
+    rove_connector: Arc<rove_connector::Connector>,
+    qc_pipelines: Arc<HashMap<(i32, RelativeDuration), rove::Pipeline>>,
```
jo-asplin-met-no, line 83: Somehow I feel that grasping the idea of pipelines is a key to understanding how rove works.

ingestion/src/kldata.rs
Lun4m: A2: note that this part is specifically about quality control (QC) of incoming data, not about storage. All incoming data will be stored in LARD, but it might not make sense to immediately QC data that comes without a set time resolution.

ingestion/src/lib.rs
```diff
@@ -93,26 +101,71 @@ impl FromRef<IngestorState> for Arc<RwLock<(ParamPermitTable, StationPermitTable
...
+    fn from_ref(state: &IngestorState) -> Arc<HashMap<(i32, RelativeDuration), rove::Pipeline>> {
+        state.qc_pipelines.clone()
+    }
+}
+
 /// Represents the different Data types observation can have
 #[derive(Debug, PartialEq)]
 pub enum ObsType<'a> {
     Scalar(f32),
     NonScalar(&'a str),
```
jo-asplin-met-no, line 120: Q1: I guess that for scalar obs values there will never be a need for storing with higher precision than f32? And if the need should arise, those should be stored as NonScalar (and we'd accept the performance penalty)?
Q2: Will there be some kind of basic validation of NonScalar observations, such as restricting to some maximum size etc.?

ingestion/src/kldata.rs
intarga: 1: Not quite, This mapping is the best we have, but a lot of type_id's don't map to a single time resolution. As time resolution is so important for end users (including ourselves in QC), we plan to transition towards time resolution being natively supported in our dataset, working first with content managers, then eventually the station network. @pmlefeuvre-met is the authority on this (and also anything stinfosys related) if you have further questions

2: There are no restrictions on time resolution in LARD, as the data table is just a heap of observations tagged with tsid and obstime. Any time resolution info would be stored outside the data table in timeseries or labels tables, and can take any form needed, as it is a requirement of our db to support irregular data, and forward compatibility with new time resolutions should be painless. This design decision is inherited from both Kvalobs and ODA. I'm open to ideas about the design or alternatives, we can discuss in the office tomorrow if you wish

db/partitions_generated.sql
intarga: Rebalancing the parititions is a costly process that requires rebuilding the whole table. Since our migration import takes a matter of hours, it's not a big deal now, but it's not something we would want to do often in production.

This is why I mentioned `obstime` being a "predictable" partition key though. Since we will only be ingesting new data in the latest partition, all other partitions should stay the correct size and never need rebalancing.

Creating new partitions and adjusting their size as our number of stations and the time resolutions they report changes will be an ongoing maintenance task though

ingestion/src/kldata.rs
intarga: The ingestor is designed to be as source-indifferent as possible.

We have generic types like `Datum` and `DataChunk`, and some functions that can QC and insert them into the DB.

To support a new data source, you do only the source specific things:

1. Define a way to receive data (i.e a new HTTP endpoint, a Kafka consumer, etc.).
2. Provide a parser for the format you receive (This is where source-specific types like `ObsinnObs` and `ObsinnChunk` may come in).
3. Provide some code to fetch or create source specific labels for the data.
4. Finish parsing by providing the data in the form of the generic types mentioned above.

In the file you're commenting on you are seeing Obsinn-specific types because that file contains the parser for Obsinn's message format "kldata".

ingestion/src/kldata.rs
intarga: Consensus is that the best way to do this is to have normal timeseries for the latitude and longitude. This is also how the stations treat it, with lat and lon being reported as params. A user would look up a stations location information, see it marked as moving, and so look up its lat and lon params too along with the data they're interested in. APIs like Frost can provide an abstraction over this as appropriate.

ingestion/src/lib.rs
intarga: 1: After investigating the performance implications, we actually decided to switch to f64 everywhere. See #54 for more info.

2: The point is that we don't know anything about the non-scalar parameters apart from that they probably shouldn't be stored as floats, so strings are the safest option to avoid data loss, as they are the form the data is delivered in. As we identify these parameters, we plan to store them in more appropriate types, i.e integer, enum, datetime; I think that's where validation becomes relevant.

ingestion/src/lib.rs
intarga: Yes, I would say so.

ingestion/src/kldata.rs
intarga: Alright, I guess we leave it unchanged then.

ingestion/src/kldata.rs
jo-asplin-met-no: Design makes sense to me. I actually have a relevant case. For this frost-ingest experiment I'm working on (designed for letting e.g. Avinor upload sets of observations as XML via the /put endpoint in a Frost instance), one option is that Frost passes those observations (decoded from the XML data) directly to lard via the lard ingestor. Will I then have two options: 1) format the data exactly like Obsinn does (so that the ingestor can't tell the difference), or 2) extend the ingestor with a parser for a new format? And then it boils down to who will be doing the translation between Avinor names and MET names; who will decide what type IDs and param IDs to use etc. Do you have any opinions on whether 1) or 2) would be the best choice in this case?

intarga: That can work. Whether 1 or 2 is best will depend on the format you receive the data in but most probably 2 is best.

Make an issue to track this, and we can discuss in more detail tomorrow.

ingestion/src/kldata.rs
jo-asplin-met-no: Does this mean that, for consistency, other obstime-dependent metadata, like quality info are also represented as normal time series? (haven't really checked the db schema, I must admit!) This is fine by me. I guess it does mean that you save space (to the extent that NULLs occupy space) for stationary time series that don't keep changing positions. Maybe some database queries for moving time series would be a little faster if the positions were in the same table as the obs time, but I don't know. Btw, are there any plans of supporting geo search directly in lard (via e.g. PostGIS)?

ingestion/src/kldata.rs
pmlefeuvre-met: Q1. Stinfosys currently maintains a default time resolution based on the typeid (see ref. below). **However, the typeid concept have not been consistently used across the station networks we receive data from.** I would strongly support that we have a time resolution attached to time series as we also transition for reducing the different usages of typeids to specifically differentiate sources of data on a substation level. 

`time_resolution` should be derived directly from the data sampling frequency or an estimate delivered by the station maintainer/responsible and as set in the logger. It can be done in a semi-automatic way, the difficulty I foresee is more when a time series change time resolution should it have a different primary key. 

For reference, I describe three tables that contain time resolution in some ways:
A. The table `time_resolution_default` in stinfosys is/was used by ODA to define a time_resolution for a `message_formatid` AND `paramid`.
B. The table `message_format` used by kvalobs, Obsinn, KDVH defines a `frequency` for a `message_formatid` only.
C. obs_pgm also has a column `hour` and `anytime` where the first defines a time_resolution on a hourly (or lower) and the second a time resolution higher than hourly.  The table is used by kvalobs and kro.

ingestion/src/kldata.rs
jo-asplin-met-no: Made [issue](https://github.com/metno/lard/issues/59).

ingestion/src/kldata.rs
jo-asplin-met-no: By quality info I mean a one-digit quality code, control info flags (16 digit) from kvalobs, etc. (this is in contrast to performance- and exposure (siting class) info that only rarely change and hence don't need to be kept with the observations). I guess the quality info I'm thinking of is (more or less) what is produced by rove.

If a change of location means a new time series, does that mean that lard makes a distinction between the station location and the obs. location? And that the station location is optional? (think of a sensor mounted in a car - then it doesn't really make sense to keep a station location at all - and as a consequence, there is always only one time series of observations to be generated from that car).

intarga: > By quality info I mean a one-digit quality code, control info flags (16 digit) from kvalobs,

This is stored alongside the data (`flags.kvdata` table for kvalobs which we will use for the beta, regular data table with the `qc_usable` column for confident)

> If a change of location means a new time series, does that mean that lard makes a distinction between the station location and the obs. location? And that the station location is optional?

We don't currently store any station-level information, we have only the timeseries.

> think of a sensor mounted in a car - then it doesn't really make sense to keep a station location at all

A moving timeseries does not have any location info for the timeseries, location is defined by the lat lon parameters, as discussed above.

jo-asplin-met-no (approved):

PR #66, opened 2025-02-10
Graceful shutdown

jo-asplin-met-no:


Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
   "ingestion",
   "integration_tests",
   "rove_connector", "util",
+  "common",
```
intarga, line 8: We already have the util crate we can reuse for this, rather than creating a new one

common/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+    tokio::select! {
+        _ = ctrl_c => {},
+        _ = sighup => {},
+        _ = sigquit => {},
+        _ = sigpipe => {},
+        _ = sigalrm => {},
+        _ = sigterm => {},
+        _ = sigchld => {},
+    }
+}
```
intarga, line 82: I'm not sure we need to listen for all these different signals. `SIGTERM` is the only relevant one, as that's what systemd sends

intarga (changes requested):

Cargo.toml
jo-asplin-met-no: I agree, will do! (I actually considered putting it under util first, but then I changed my mind for some reason)

common/src/lib.rs
jo-asplin-met-no: Ok, will reduce it to SIGTERM and SIGINT (Ctrl-C) then. But do you think we should only catch SIGTERM? Hm ... I guess I don't see exactly what use case we are trying to support here. In particular: what exactly do we risk by not catching given signals? Could it for example result in some kind of inconsistency/incorrectness? And/or that processes stick around (allocating resources) when they shouldn't? (not trying to arguing against catching signals here, only interested in understanding things :) )

api/src/lib.rs
```diff
@@ -135,5 +135,8 @@ pub async fn run(pool: PgConnectionPool) {
 
     // run it with hyper on localhost:3000
     let listener = tokio::net::TcpListener::bind("0.0.0.0:3000").await.unwrap();
-    axum::serve(listener, app).await.unwrap();
+    axum::serve(listener, app)
+        .with_graceful_shutdown(::util::shutdown_signal())
```
intarga, line 139: The leading scoping operator before `util` looks a little odd here, is it correct?

util/src/lib.rs
```diff
@@ -1 +1,27 @@
...
+
+    #[cfg(unix)]
+    let sigterm = async {
+        signal::unix::signal(signal::unix::SignalKind::terminate())
+            .expect("failed to install signal handler for SIGTERM")
+            .recv()
+            .await;
+    };
+    #[cfg(not(unix))]
+    let sigterm = std::future::pending::<()>();
```
Lun4m, line 19: Do we even need the `cfg`s here? I don't think we will ever want to deploy on Windows?

```suggestion
    let sigterm = async {
        signal::unix::signal(signal::unix::SignalKind::terminate())
            .expect("failed to install signal handler for SIGTERM")
            .recv()
            .await;
    };
```

common/src/lib.rs
jo-asplin-met-no: I agree - all good points here. I guess I was kind of hoping/expecting to get some feedback of this kind on this change so I got a chance to learn something. "Systems level" programming considerations like these I encounter only rarely, so it's good to get a reminder (and believe it or not, but the concept of "systemd process management" was new to me (I know what  process is and that you can interrupt it with a signal)). And I absolutely agree that we shouldn't throw in a bunch of signals just in case. Will reduce it to just SIGTERM then, and add a note about why.

util/src/lib.rs
jo-asplin-met-no: Good point. We shouldn't pretend that the code is portable if we don't intend for it to be. It could in fact be misleading!

intarga: > I don't think we will ever want to deploy on Windows?

It's possible in the future if other institutes adopt our codebase there might be demand to make it work on Windows. I agree it's not worth considering for now though, we can cross that bridge if and when we come to it, hopefully with some funding for a windows CI runner

common/src/lib.rs
intarga: To be clear about what systemd does:

When stopping a process (with i.e. `systemctl stop` or `systemctl restart`) it sends `SIGTERM` immediately, followed by `SIGKILL` after 90 seconds. The intention is to give programs a short window to gracefully shutdown, then forcefully terminate if they seem to be hanging.

jo-asplin-met-no: Since we're on the subject: Do you have an idea of what could have happened if we didn't catch SIGTERM and passed the fact to with_graceful_shutdown()? Some kind of resource leakage?

intarga: Data corruption is the concern. Since we do our SQL queries in transactions, the worst data corruption that should be mitigated anyway, but this furthur helps us towards the ideal case where a request from obsinn is either processed in it's entirety and receives a success response, or doesn't change the DB at all.

intarga:
It's just occurred to me that we should probably handle the kafka task spawned at ingestion/src/main.rs:69

The stinfo task spawned just before it should be fine to leave unhandled though.

common/src/lib.rs
intarga: > considerations like these I encounter only rarely

Actually I think this is relevant to other services you're responsible for, like Frost on Kubernetes. Kubernetes handles termination the same way systemd does with `SIGTERM` followed by `SIGKILL`, albeit with a shorter default grace period of 30 seconds:

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination

jo-asplin-met-no: Yup. In Frost we do implement graceful shutdown by handling these: SIGTERM, SIGINT, SIGHUP, and SIGQUIT. I realize now that the latter two didn't need to be there.

api/src/lib.rs
jo-asplin-met-no: The leading scoping operator needs to be there for the code to compile.

jo-asplin-met-no:
> It's just occurred to me that we should probably handle the kafka task spawned at ingestion/src/main.rs:69
> 
> The stinfo task spawned just before it should be fine to leave unhandled though.

I'll have a go at implementing graceful shutdown for this task then (in this PR if that's ok - shout out if you think it should be in a separate issue/branch/PR). I think I can apply the more low-level idiom I tested last week.

ingestion/src/kvkafka.rs
```diff
@@ -245,28 +250,36 @@ async fn read_kafka(group_name: String, tx: mpsc::Sender<Msg>) {
...
+                            .commit_consumed()
+                            .expect("could not commit offset in consumer"); // ensure we keep offset
                     }
-                    if let Err(e) = consumer.consume_messageset(msgset) {
-                        eprintln!("{}", e);
+                    Err(e) => {
+                        eprintln!("{}\nRetrying in 5 seconds...", Error::Kafka(e));
+                        tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
                     }
                 }
```
intarga, line 281: This select does not do what you think it does. Instead of telling you why I think you will learn more if you figure it out yourself, so I will ask a leading question instead: Can this loop break while only halfway through processing a message?

Feel free to ask follow on questions if you're stuck

intarga (changes requested):

ingestion/src/main.rs
```diff
@@ -66,25 +68,44 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+        Ok(_result) => {}
+        Err(_error) => {}
+    };
+
+    // At this point the ingestion server has finished (possibly due to a shutdown signal).
+    // Ask other tasks to clean up and finish too:
+    cancel_token.cancel();
+    tracker.wait().await;
+
+    Ok(())
```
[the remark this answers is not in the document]
jo-asplin-met-no: Well, if the other tasks (ingestor and kvkafka reader) all terminated for some other reason than a signal being caught, then the signal catcher will never terminate, and it would feel wrong to include it in the join. And vice versa: if the other tasks did terminate as a result of the cancellation token being triggered by the signal catcher, then we know that the latter has done it's job and we don't really care about it anymore, so no need to join it in this case either. So no, I don't think the signal catcher should be joined. Am I right?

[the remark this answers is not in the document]
jo-asplin-met-no: I guess I was wrong. I forgot that the signal catcher needs to be collected like any other task, so it must be included in the join. The body of the task (the one that listens for signals) doesn't start until an await is called on the return value from spawn or that return value is passed to join (which does the equivalent of await behind the scenes I guess).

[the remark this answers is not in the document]
intarga: Your first answer was right

>  I forgot that the signal catcher needs to be collected like any other task

It doesn't need to be collected if we don't care about it's state, and if the other tasks have terminated, we no longer have a use for it.

> The body of the task (the one that listens for signals) doesn't start until an await is called on the return value from spawn or that return value is passed to join

This is true of futures, not tasks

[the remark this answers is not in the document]
jo-asplin-met-no: (thanks for putting up with all my questions btw - but rest assured that this type of feedback is really helpful for me!)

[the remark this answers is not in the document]
intarga: No worries! My framing of "What do you think?" was inviting a more interactive discussion :slightly_smiling_face:

ingestion/src/kvkafka.rs
jo-asplin-met-no: FYI: I still haven't dealt with this. Will take a look tonight. Got to rush home now.

jo-asplin-met-no: Pushed another attempt now.

jo-asplin-met-no: I guess the above cannot break in the middle of processing a message, and so there's no point in checking the cancellation token in that loop (thus the second attempt at a solution which reverted the loop to its original form). I guess the essential point is that there are two types of signals that can happen: 1) SIGKILL and 2) the rest. SIGKILL can happen absolutely anywhere and at all times, but it can also not be caught, so there's no point in trying to handle it (it's like pulling the power chord of a traditional (non-battery) computer). All rest of the signals can be caught, but that can only happen in the task where you have explicitly registered a handler for it, i.e. other tasks (like the one running the above loop) can safely assume that the signal doesn't pop up there as well. That's my current view of the world.

intarga: >I guess the above loop cannot break in the middle of processing a message

Wrong. Maybe it will be easier to spot why in a golang equivalent:
```
msgset_processed := make(chan bool, 1)

go func() {
    for {
        let msgset = consumer.poll()
    
        handle_msgset(msgset) // parsing + database transactions + marking offset
    
        msgset_processed <- true
    }
}()

for {
    select {
    case <-cancellation_channel:
        break
    case <-msgset_processed:
        continue
    }
}
```

Can this break partway through processing a message?

In chat my suggestion to you was:
```
loop {
    select! {
        _ = shutdown_channel.recv() => { break; }
        poll_result = consumer.poll() => {
            // rest of the body here ...
        }
    }
}
```

Which is equivalent to this golang:
```
msgsets := make(chan Msgset, 1)

go func() {
    for {
        let msgset = consumer.poll()
    
        msgsets <- msgset
    }
}()

for {
    select {
    case <-cancellation_channel:
        break
    case <-msgset_processed:
        handle_msgset(msgset)
    }
}
```

Do you understand why this is different?

jo-asplin-met-no: Hm ... well, yes I understand that you can listen for the cancellation in that loop by using select!. But I guess I don't understand why you have to when there's already a task that does that (the signal_catcher spawned in ingestion/main).

intarga: This isn't about signals at all, consider `cancellation_channel`/`shutdown_channel` to be your cancellation token. There's a basic concurrency bug lurking in your version.

intarga: > Why isn't it sufficient to handle the cancellation in the read_and_insert function? In my latest attempt there's now a select! inside an infinite loop that selects between await_cancellation and insert_kvdata.

Please stay on topic and make sure you understand the first problem before throwing new changes around.

> Note also that my implementation is based solely on using a cancellation token. I haven't considered using a channel like you suggest (shutdown_channel.recv()). Would that be better practice?

Not relevant to the conversation, they're equivalent for this purpose. I wrote that snippet before you started using a cancellation token.

Please ignore specifics, the bug is a lot more fundamental. Considering the golang snippets here should be enough.

jo-asplin-met-no: Is the point here the calls to .await in the body B that is run once consumer.poll() returns (e.g. parse_message(...).await in the Ok() case)? I guess the tokio runtime could let another task run at that point, but the question is then how can that other task result in something that would result in B being broken prematurely? (except for SIGKILL of course).

You say that the key can be found just by looking at the golang snippets, but I'm afraid I'm unable to see it. The difference seems to be that in the first case handle_msgset is run in a separate task/goroutine from the one that catches the cancellation, while in second case those two events are run in the same task/goroutine (i.e. the main task). Is the solution here that the latter case guarantees that the cancellation can't interfere with handle_msgset?

Lun4m: Hopefully I'm not giving away too much, but I guess the key here is understanding how `select` works (I've also been bitten by this before :)

intarga: In the first go snippet, what happens if a value appears on `cancellation_channel` while `handle_msgset` is running? The main goroutine is waiting for a value on either channel, and `msgset_processed` is empty, so it pulls the value off `cancellation channel`, breaks the loop and ends the process, dropping the other goroutine breaking off its execution in the middle of `handle_msgset`.

This does not happen in the second snippet, because when `handle_msgset` is running, `cancellation_channel` is not being waited on.

jo-asplin-met-no: Ok, saved by the bell there I guess. I'll try yet another attempt then. I think I understand the point with the golang snippets, and I'll try to transfer that to rust/tokio based on your suggested sketch. But apparently the only difference from that sketch and my original attempt was that the latter had an async in front of the message handling branch body, i.e.

```
loop {
    select! {
        _ = shutdown_channel.recv() => { break; }
        _ = async {
            match consumer.poll() => {
            // rest of the body here ...
            }
        }
    }
}
```
instead of

```
loop {
    select! {
        _ = shutdown_channel.recv() => { break; }
        poll_result = consumer.poll() => {
            // rest of the body here ...
        }
    }
}
```

So maybe my original attempt would be fine after if only I got rid of async? (don't remember exactly why I put it there)

jo-asplin-met-no: But my original attempt looks fishy anyway since it has no => at the uppermost level in the select! construct.

intarga: >  But apparently the only difference from that sketch and my original attempt was that the latter had an async in front of the message handling branch body.

That is not the only difference. Look again.

> And my original attempt looks fishy anyway since it has no => for the message handling branch at the uppermost level in the select! construct.

It does, it's just cut off above

intarga: In my version I only select against the future returned by consumer.poll(). In your version, you create a future for the whole loop body and select against that.

Lun4m: Unfortunately the kafka library we are using is not async, that's why you need the async block (in either version).

jo-asplin-met-no: Yet another attempt now. This time more aligned with @intarga 's suggestion. I did need to wrap consumer.poll() in an async block. Will that work?

ingestion/src/kvkafka.rs
```diff
@@ -229,27 +234,35 @@ async fn read_kafka(group_name: String, tx: mpsc::Sender<Msg>) {
...
+                    Ok(sets) => {
+                        for msgset in sets.iter() {
+                            for msg in msgset.messages() {
+                                if let Err(e) = parse_message(msg.value, &tx).await {
+                                    eprintln!("{}", e);
+                                }
+                            }
+                            if let Err(e) = consumer.consume_messageset(msgset) {
+                                eprintln!("{}", e);
+                            }
```
Lun4m, line 255: Another option with cancellation tokens (which I've never used before by the way) is to skip the `select` altogether.
I guess it's a matter of preference and it doesn't do exactly what the select code does (in particular, it doesn't play well with the `Err` branch of the match statement), but I find it easier to reason about.

```suggestion
        if cancel_token.is_cancelled() {
            eprintln!("cancellation token triggered");
            break;
        }

        // https://docs.rs/kafka/latest/src/kafka/consumer/mod.rs.html#155
        // poll asks for next available chunk of data as a MessageSet
        match consumer.poll() {
            // process message sets
            ...
        }
```

ingestion/src/kvkafka.rs
intarga: This has resolved the race condition, yes. The important thing here though is that you understand why there was a race condition here, and feel confident you can spot race conditions in the future.

Every project you work on at met is highly concurrent. Race conditions are a constant danger in such an environment, and they often come intertwined with complicated code they're hard to untangle from. If you can't see a race condition in a minimal example, you will be cursed to constantly write spooky bugs in your concurrent code. I would not be happy leaving it there if I were you.

intarga: There's another concurrency bug (not introduced by you) in the kafka code that I've noticed in this review, but I don't mind if you want to leave that to be fixed in a later PR, as this one has already been quite a lot.

ingestion/src/kvkafka.rs
intarga: Yes, I forgot that our kafka library is sync. In that case I think this is equivalent to the select, and probably the way to go. It has another problem though: In the pathological case where the kafka queue is empty when we cancel, This will hang indefinitely (or at least the 90 seconds until systemd sends `SIGKILL`...). It's also a problem that we're doing blocking IO on a non-blocking task.

We can find a way around this, but I'm starting to think it would make more sense to switch to `rdkafka` which supports async.

Lun4m: Why indefinitely? If the queue is empty I'd expect `consumer.poll()` to return an error (?), after which we wait 5 seconds and check if the token was cancelled before polling again.
But maybe it's not so simple and I agree we should probably switch to `rdkafka`.

> It's also a problem that we're doing blocking IO on a non-blocking task.

This is the bug you mentioned in the other comment? That we are simply using `spawn` instead of `spawn_blocking`?

ingestion/src/kvkafka.rs
jo-asplin-met-no: > This has resolved the race condition, yes. The important thing here though is that you understand why there was a race condition here, and feel confident you can spot race conditions in the future.

So to summarize a bit (so that I can actually learn from this exercise and be able to spot possible race conditions in the future (which I of course really do appreciate the importance of)), the difference between my original proposal and the one I ended up with seems to be quite subtle.

The original/erroneous proposal was essentially this:

```
    // Consume the kafka queue infinitely
    loop {
        tokio::select! {
            _ = cancel_token.cancelled() => { break }
            _ = async { match consumer.poll() { /* body */ } => {}
        }
    }
```

while the final/correct proposal is essentially this:

```
    // Consume the kafka queue infinitely
    loop {
        tokio::select! {
            _ = cancel_token.cancelled() => { break }
            poll_result = async { consumer.poll() } => { match poll_result { /* body */ }
        }
    }
```
But I do think I now (finally) see the essence of the problem:
- The select! statement starts multiple _computations_ concurrently.
- A _computation_ here is the statement(s) _before_ the '=>'. (I missed this point in my first proposal!)
- When the first computation completes, two things happens:
  1: all the other computations are dropped immediately.
  2: the statement(s) that follows the '=>' of the completed computation is/are executed.
  3: the select! statement as such completes
  
So having _both_ the consumer.poll() _and_ the message processing body _before_ the '=>' was really the mistake, since the body could be dropped partway through if the other computation that listened for a cancellation suddenly completed!

The assumption here is of course that the computation inside consumer.poll() is not sensitive to being dropped at any time (at least not at a level that matters to us).

intarga: Yes! Nicely expressed

intarga: > all the other computations are dropped immediately.

Minor correction, they aren't immediately dropped, the futures are just left in their current state (guaranteed to be `Pending`). They're dropped when the function returns and they go out of scope, which is in this case almost immediately.

ingestion/src/kvkafka.rs
Lun4m: > Nope. What I was talking about is that at the moment we commit offsets without waiting for the associated DB inserts to complete. Related to that the DB inserts being on a separate task means they aren't covered by the graceful shutdown, but a solution to the first problem with probably also solve the second.

I think it would be covered indirectly by the sender being dropped? But oof. good catch, it probably makes more sense to parallelize over message sets instead of single messages.

jo-asplin-met-no:
FYI: I now did some manual testing of the `axum::serve(...).with_graceful_shutdown(util::await_cancellation(cancel_token))`. I followed the suggestion of @intarga and made a temporary endpoint that prints 'sleeping for 20 secs', then sleeps for 20 secs, and finally returns some 200 Ok response. Then I called this endpoint from curl and quickly after sent a signal to the pid with `kill ...`. Watching the behavior for different combinations of signals and including/excluding `.with_graceful_shutdown(...)` gave the expected results. In particular, sending SIGTERM with `.with_graceful_shutdown(...)` included causes the endpoint handler to complete (returning the expected 200 Ok response) and the process then to terminate.

util/src/lib.rs
```diff
@@ -1 +1,34 @@
...
+        _ = ctrl_c => {},
+    }
+
+    cancel_token.cancel()
+}
+
+/// Returns a Future that completes once cancel_token is triggered.
+pub async fn await_cancellation(cancel_token: CancellationToken) {
+    _ = cancel_token.cancelled().await;
+}
```
intarga, line 34: This function seems unnecessary

ingestion/src/main.rs
```diff
@@ -66,25 +67,41 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+        (_, _) = (ingestor_res, kvkafka_reader_res); // ignore for now
+    }
+
+    #[cfg(not(feature = "kafka_prod"))]
+    {
+        let ingestor_res = tokio::join!(ingestor);
+        _ = ingestor_res; // ignore for now
+    }
+
+    Ok(())
```
intarga, line 106: This has changed from returning a result from the main function to ignoring it. Not a big deal, but since the results are there I think we should use them.

In the `kafka_prod` case you can combine the two results with `.and_then()`

api/src/lib.rs
jo-asplin-met-no: Done!

util/src/lib.rs
jo-asplin-met-no: Agreed. There was some technical reason why I wrapped it. I'll take another look and see if I can solve the problem in a simpler way.

ingestion/src/main.rs
jo-asplin-met-no: I'll take a look (after lunch).

util/src/lib.rs
jo-asplin-met-no: Done!

ingestion/src/main.rs
```diff
@@ -66,25 +67,41 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+    #[cfg(feature = "kafka_prod")]
+    // Spawn kvkafka reader
+    {
+        let kafka_group = args[1].to_string();
+        println!("Spawning kvkafka reader...");
+        let kvkafka_reader = tokio::spawn(lard_ingestion::kvkafka::read_and_insert(
+            db_pool.clone(),
+            kafka_group,
+            cancel_token.clone(),
+        ));
```
Lun4m, line 94: Need to switch around the `clone()` calls (for `db_pool` and `cancel_token`). I guess we are not building with the `kafka_prod` feature in the CI pipeline (maybe we should), so it compiles fine :sweat:

jo-asplin-met-no: Done!

ingestion/src/main.rs
jo-asplin-met-no: Done!

ingestion/src/main.rs
```diff
@@ -66,25 +67,36 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+        let kafka_group = args[1].to_string();
+        println!("Spawning kvkafka reader...");
+        let kvkafka_reader = tokio::spawn(lard_ingestion::kvkafka::read_and_insert(
+            db_pool,
+            kafka_group,
+            cancel_token,
+        ));
+
+        let (ingestor_res, kvkafka_reader_res) = tokio::join!(ingestor, kvkafka_reader);
+        ingestor_res.unwrap().map(|_| kvkafka_reader_res.unwrap())
```
intarga, line 97: Not what I meant. Here you panic on error instead of returning it

jo-asplin-met-no: I thought it didn't matter since the program was exiting at this point anyway, and I assumed the details would end up in some log (stderr?). I actually tried the .and_then() idiom first, but ran into some problems. I'll take another look.

jo-asplin-met-no: (and I guess I shouldn't just assume things, but actually test it ;) )

Lun4m: Isn't it panicking on panic, since `join` returns `Result<T, JoinError>`? The actual error should be inside `T`, no? But I guess we can still use question marks instead, and then you wouldn't need to use the map function.

jo-asplin-met-no: Done!

jo-asplin-met-no:
@intarga @Lun4m: So now I finally tested graceful shutdown of the kvkafka message processing in a similar way as I did for the axum endpoint. Here's what I did: 1) added printouts right after the successful reception of messages from consumer.poll(), 2) added a 5 sec sleep after the printout, 3) added another printout right after the corresponding DB commit, 4) signaled the process with `kill -15` and confirmed the expected presence/absence of second printout depending on whether or not the handling of SIGTERM was commented out.

If you think this suffices as a test for now, please give a thumbs up and I'll merge the PR.

Here's the diff of the test code:

```
diff --git a/ingestion/src/kvkafka.rs b/ingestion/src/kvkafka.rs
index 81f103a..cdc7a5c 100644
--- a/ingestion/src/kvkafka.rs
+++ b/ingestion/src/kvkafka.rs
@@ -232,6 +232,9 @@ async fn read_kafka(group_name: String, tx: mpsc::Sender<Msg>, cancel_token: Can
     .create()
     .expect("failed to create consumer");
 
+    println!("created kvkafka reader");
+    let mut i = 0;
+
     // Consume the kafka queue infinitely
     loop {
         tokio::select! {
@@ -244,6 +247,9 @@ async fn read_kafka(group_name: String, tx: mpsc::Sender<Msg>, cancel_token: Can
             poll_result = async { consumer.poll() } => {
                 match poll_result {
                     Ok(sets) => {
+                        i += 1;
+                        println!("got MessageSets ({i}); sleeping for 5 secs ...");
+                        tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
                         for msgset in sets.iter() {
                             for msg in msgset.messages() {
                                 if let Err(e) = parse_message(msg.value, &tx).await {
@@ -257,6 +263,7 @@ async fn read_kafka(group_name: String, tx: mpsc::Sender<Msg>, cancel_token: Can
                         consumer
                             .commit_consumed()
                             .expect("could not commit offset in consumer"); // ensure we keep offset
+                        println!("    committed MessageSets ({i})");
                     }
                     Err(e) => {
                         eprintln!("{}\nRetrying in 5 seconds...", Error::Kafka(e));
```

PR #77, opened 2025-03-06
Prerelease documentation

intarga:
This PR includes

- A `README` for the project
- A set of markdown documents in /docs covering the architecture of lard's components (with the exception of api [now egress] because its design is not mature yet and won't be before the beta)
- Renaming of `api` to `egress` as `api` is a rather redundant and non-descriptive name, especially since `ingestion` technically contains an API. I feel the name `egress` better reflects its purpose.
- Removal of `fake_data_generator` as it is disused and not really maintained


docs/DATABASE.md
```diff
@@ -0,0 +1,64 @@
+# Database
+
+The database is defined by a set of schemas defined in [/db](/db) each prefixed by a number defining the order in which they should be applied.
```
Lun4m, line 3:
```suggestion
The database is defined by a set of schemas found in [/db](/db), each prefixed by a number representing the order in which they should be applied.
```

docs/INGESTION.md
```diff
@@ -0,0 +1,99 @@
...
+                retry: !matches!(e, Error::Parse(_)),
+            })
+        }
+    }
+}
+```
+
+## Kafka and legacy
+
+Right now as we aren't ready with confident QC, we are ingesting from KvKafka into the `legacy.data` table that has a different structure. As a result, that ingestor has it's own equivalent to `Datum` and `insert_data` that match `legacy.data`.
```
Lun4m, line 99:
```suggestion
Right now as we aren't ready with confident QC, we are ingesting from KvKafka into the `legacy.data` table that has a different structure. As a result, that ingestor has its own equivalent to `Datum` and `insert_data` that match `legacy.data`.
```

docs/INGESTION.md
```diff
@@ -0,0 +1,99 @@
...
+            Json(KldataResp {
+                message: e.to_string(),
+                message_id: 0, // TODO: some clever way to get the message id still if possible?
+                res: 1,
+                retry: !matches!(e, Error::Parse(_)),
+            })
+        }
+    }
+}
+```
```
Lun4m, line 95: Should this just be a link to the lines in the file? We have to fix it here too when the underlying code changes? But I also kinda like seeing the code directly in the README :thinking:

docs/INTEGRATION_TESTS.md
```diff
@@ -0,0 +1,96 @@
...
+            period: Duration::hours(1),
+            type_id: 501,
+            len: 48,
+        };
+
+        // As ingestion and egress are exposed as HTTP servers, we need an HTTP client to interact
+        // with them
+        let client = reqwest::Client::new();
+
+        // Use helper fucntion `ingest_data` to send a request with the obsinn message generated
```
Lun4m, line 29:
```suggestion
        // Use helper function `ingest_data` to send a request with the obsinn message generated
```

docs/INTEGRATION_TESTS.md
```diff
@@ -0,0 +1,96 @@
...
+                    "TRUNCATE public.timeseries, labels.met, labels.obsinn CASCADE",
+                )
+                .await
+                .unwrap();
+        }
+        assert!(test_result.is_ok())
+    }
+}
+```
+Here we expect the test_result to complete first, as the server tasks should run indefinitely unless they crash or until we tell them to shut down, so we panic and fail the test if this does not happen.
```
Lun4m, line 92:
```suggestion
Here we expect the `test_result` to complete first, as the server tasks should run indefinitely unless they crash or until we tell them to shut down, so we panic and fail the test if this does not happen.
```

docs/INTEGRATION_TESTS.md
```diff
@@ -0,0 +1,96 @@
...
+
+In the handler for the test closure, we make sure to clean up the database before moving on, so the next test can start on a blank slate. We make sure to catch any panics so we will perform this cleanup even if the closure fails.
+
+It's worth noting that e2e_test_wrapper does not set up a postgres instance, just connects to it, so a postgres instance must be available to run these tests. To help with this, we've set up a justfile to use instead of running `cargo test` directly, which sets up a postgres instance in a docker container.
```
Lun4m, line 96:
```suggestion
It's worth noting that `e2e_test_wrapper` does not set up a postgres instance, just connects to it, so a postgres instance must be available to run these tests. To help with this, we've set up a justfile to use instead of running `cargo test` directly, which sets up a postgres instance in a docker container.
```

Lun4m:
Also, maybe we can remove the `README.md` inside the `integration_tests` directory?

README.md
```diff
@@ -1,56 +1,55 @@
...
+
+## Status
+
+Approaching beta, targeting summer 2025.
+
+## Architecture
+
+Lard is built around a Postgres database with two services that interact with it, one focused on ingestion, and one providing an API to access the data.
+
+![Diagram of the architecture on a single node](/docs/images/single-arch.svg)
```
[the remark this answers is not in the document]
intarga: I don't like having them in one image because I think there's value in presenting them separately, and I think they would be too cluttered in one image.

I thought about having the single version horizontal, but I didn't like that it broke symmetry with the multi version

README.md
```diff
@@ -1,56 +1,55 @@
...
+
+Here, one node takes responsiblity for ingestion, using [Postgres replication](https://www.postgresql.org/docs/current/high-availability.html) to sync the others. Meanwhile, the others focus on serving read-only requests from the API service, allowing read throughput to scale linearly with the number of replicas. Replicas are also able to take over from the primary in case of outages, minimising downtime.
+
+In addition to read throughput, previous experience with database systems at Met has taught us that as our dataset grows (think past 1 billion observations) write throughput begins to slow to a problematic degree. This happens because the [indexes](https://www.postgresql.org/docs/current/indexes.html) (structures needed speed up queries on large tables) become resource intensive to maintain as they grow larger. Particularly the BTree indices we use to represent time need to remain a balanced, but as we always add data on one side of the tree (the present is one extreme of the time range our dataset covers), we are constantly unbalancing it, and the expense of balancing a tree scales with its size.
```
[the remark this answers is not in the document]
intarga: Slow is grammatically correct.

I'm not sure either about Met vs MET, I think I've seen it written both ways, idk if there's a style guide. Personally I think Met makes more sense though since it's not an acronym

docs/INGESTION.md
intarga: I think linking the lines will be too fragile, IMO it's a better failure mode to have the snippet be outdated than incomprehensible

README.md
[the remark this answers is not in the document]
intarga: > Can the SVGs have a solid white background? They are unreadable in dark mode 😢

Did one better, now there are dark mode versions of the images 😁

README.md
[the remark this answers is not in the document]
intarga: Fixed the other two typos

docs/DATABASE.md
intarga: Fixed

docs/INTEGRATION_TESTS.md
intarga: Fixed

docs/INTEGRATION_TESTS.md
intarga: Fixed

docs/INTEGRATION_TESTS.md
intarga: Fixed

Lun4m (approved):
Beautiful!

PR #93, opened 2025-04-02
Update rove

intarga:
Build issues in CI were being caused by a transitive dependency of rove. This should fix them


Lun4m:
It's still using `Compiling faer v0.21.4`?

intarga:
> It's still using `Compiling faer v0.21.4`

It's the lock file

PR #94, opened 2025-04-02
Remove cargo check and update yamlfmt in CI

Lun4m:


Lun4m:
`cargo check` (+ `clippy` that depends on it) and `cargo build` don't use the same compilation metadata, so they both need to compile the crates "from scratch".

Issue #67, opened 2025-02-11
Backups


intarga:
We should have some facility for regular automated backups and testing of them

PR #96, opened 2025-04-07
Create routine for backup of DB

lcoram:


lcoram:
So far have tested this on a VM with postgres installed and created the DB with a tiny bit of fake data. Will check that the crons run, then would consider that we want to run the ansible on the cluster and see that it starts taking backups. Then we should potentially use task https://github.com/metno/lard/issues/67 to check it actually works on larger amounts of data / the real cluster?

lcoram:
## Create database
`sudo -u postgres psql`
`create database lard`
enter the lard database
`sudo -u postgres psql -d lard`
apply all the .sql files from the \db folder

### Made some fake data
`insert into timeseries(id) values (123),(456);`
`insert into data(timeseries, obstime, obsvalue) values (123, '2025-01-01 10:20:30', 0), (456, '2025-01-01 10:20:30', 1);`

### This then looks like:
lard=# select * from data;
 timeseries |        obstime         | obsvalue | qc_usable 
------------+------------------------+----------+-----------
        123 | 2025-01-01 10:20:30+00 |        0 | t
        456 | 2025-01-01 10:20:30+00 |        1 | t
(2 rows)

lard=# select * from timeseries;
 id  | fromtime | totime | loc | permit | deactivated 
-----+----------+--------+-----+--------+-------------
 123 |          |        |     |        | 
 456 |          |        |     |        | 
(2 rows)

## Create a backup to the S3 the same way the cron should
`sudo -u postgres pg_dump lard | s3cmd put - s3://lard/backups/lard__$(date +%Y%m%d%H%M%S)`
`sudo -u postgres pg_dumpall --globals-only | s3cmd put - s3://lard/backups/globals__$(date +%Y%m%d%H%M%S)`

Currently a bit unclear if we need the globals, and for this testing we probably don't. However, they could be useful if the postgres installation was more thoroughly wiped? 

## Delete the database
Then should be able to test *deleting* the database and recreating it
`sudo -u postgres psql`
`drop database lard`
Check its really not there... 
`sudo -u postgres psql -d lard`

## Recreate the database 
`sudo -u postgres psql`
`create database lard`
Try to get the backup back from the s3 and pipe it into the database.

## Get the data back
If doing this for real, find the most recent file!
`s3cmd get s3://lard/backups/lard__20250407131431 lard__latest`

`sudo -u postgres psql -U postgres -d lard < lard__latest`

Go in and have a look around...
`sudo -u postgres psql -d lard`

intarga (commented):
Nice work!

I think there are a few more things we should address before closing the issue (I don't mind if you want to address them in a follow-on PR instead of this one):

- **Playbook for restoring backups** - Ideally we should be able to just do `ansible-playbook ... restore_backup.yml`. I know the filename changes, but it should be straightforward to figure out which is the latest programatically.
- **`lard_restricted` backups** - This seems to just backup/restore the `lard` db, but it should cover `lard_restricted` too.
- **Streaming restores** - I think you just need to add a `-` like you do in the put command
- **S3 space management** - I think this is the thorniest issue. If we're doing daily multi-terabyte backups, we're quickly going to run out of space on our S3 cluster; we need to find a way to not do that.

ansible/roles/backups/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+    dest: /home/ubuntu/
+    mode: "0755"
+  become: true
+
+# backup to s3 in lard/backups bucket
+- name: Create cron to backup Lard DB
+  ansible.builtin.cron:
+    name: "backup lard"
+    minute: "0"
+    hour: "3,15" # backup daily at 3am and 3pm
```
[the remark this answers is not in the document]
lcoram: I wanted to have one that was more in the night, and maybe one at the end of the work day. The problem is that its UTC, so currently 2 hours off (so actually 5 and 17) but then that changes with winter time. Peak ingestion occurs just after the hour and then usage probably peaks around 10-15 after the hour... So maybe better to do it at like 45 over something?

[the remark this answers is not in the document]
intarga: Hmm is there a particular reason we want a 2x daily schedule? I'd have thought 1x daily was good enough since we have a comfortably larger buffer than that on the kafka queue / Obsinn. Only doing it at night would make it easier to find a good schedule

lcoram:
Agree on all points @intarga. I will test if I can do a streaming restore (although worried if its large and gets a network glitch while it's happening it could cause issues). I will also look more into incremental backups and some sort of thinning cron job for full backups. Additionally I think I need the name of the VM in the backup... Since we backup both a and b (and one could be corrupt or something)!

lcoram:
We are currently using postgres 16 right? We might have to go to 17 to get incremental backups... Seems it is a new feature.
https://www.postgresql.org/docs/current/continuous-archiving.html#BACKUP-INCREMENTAL-BACKUP

ansible/roles/backups/tasks/main.yml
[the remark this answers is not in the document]
lcoram: Ok, then just sometime during the night. At least for a full backup (that we then thin out)... We could consider incremental at other times?

intarga:
> (although worried if its large and gets a network glitch while it's happening it could cause issues)

That's a good point. I guess we have to decide what we want to prioritise. Streaming has the advantage of being faster, which if we're trying to recover from a fault means less downtime.

> Additionally I think I need the name of the VM in the backup... Since we backup both a and b

I forgot this will run on both 😱 Seems kinda wasteful? The primary is the authoritative version, so if we only going to backup one it should be that. I guess it's a bit tricky for the cron job to know whether it's on the primary or not though, and that adds an extra point of failure...

intarga:
> We are currently using postgres 16 right? We might have to go to 17 to get incremental backups... Seems it is a new feature.

Nope, we are already on 17 😁

lcoram:
Good point about wasteful, maybe I can try to wrap the Cron in a check to repmgr if its primary?

ansible/roles/backups/tasks/main.yml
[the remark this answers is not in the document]
intarga: Yeah, I imagine an incremental backup is much lighter on the system, so we can probably get away with doing those whenever

Lun4m:
If switch the cron job ON/OFF on switchover/failover. it should be okay, no? Anyway incremental backups use `pg_basebackup` (dumps the whole cluster? Version upgrade incompatible?) which is different from `pg_dump`

cskarby:
On the new kubernetes cluster the postgresql-operator has support for doing database backups

ansible/roles/recovery/files/recovery_dump.sh
```diff
@@ -0,0 +1,13 @@
...
+
+echo "Starting restore of: $1 into $2"
+
+# we have 16 VCPUs on the VMs currently, so hopefuly -j 8 is ok 
+pg_restore -U postgres -d "$2" -j 8 "$1"
+
+seconds=$SECONDS
+ELAPSED="Elapsed: $((seconds / 3600))hrs $(((seconds / 60) % 60))min $((seconds % 60))sec"
+
+echo "Finished restore of: $1, took: $ELAPSED"
```
lcoram, line 13: Tested: Finished restore of: /mnt/ssd-data/lard_restricted_20250602182145.dump, took: Elapsed: 0hrs 36min 26sec

ansible/roles/backups/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+        weekday: "1" # every Monday
+        job: "/home/ubuntu/cron_scripts/lard_restricted_backup.sh"
+
+    - name: Create cron to do base backup
+      ansible.builtin.cron:
+        name: "basebackup lard"
+        minute: "0"
+        hour: "5" # backup monthly at 5:00am (UTC)
+        day: "1" # 1st of every month
+        job: "/home/ubuntu/cron_scripts/base_backup.sh"
```
intarga, line 49: Do we want this active? Given we don't have a workflow for using these?

ansible/inventory.yml
```diff
@@ -3,3 +3,6 @@ servers:
   hosts:
```
intarga, line 3: The line I'm trying to comment (3) is just cut out of the diff, so GitHub won't let me comment it. If we're adding a new inventory "staging", we should probably rename the existing one to "production"

intarga (changes requested):
I think I noted this once before, but I think it's a big mistake to have separate playbooks for production and staging because

- It makes it very easy to mess up future changes, since you have to remember to make the change in both the prod and staging playbooks, and be carefully aware of any differences
- It defeats the purpose of having staging at all. Staging is there to test updates and ops actions without risking prod, but if all the ops code and infra for staging is different than for prod, what does such a test really tell you?
- It's just more to maintain, and more fragility

I understand you did this because we have 2 nodes in prod and only 1 in staging. There's a couple of things we can do about this:

1. Make the playbooks smart enough to handle an arbitrary number of nodes in the inventory
2. Make staging have multiple nodes

Ultimately it would be good to have both, but for now I think just doing (2) is the pragmatic approach. What do you think?

lcoram:
I agree, but I really would prefer to look into how to make it so I can start the notebook with different amounts of nodes. You are right it might be easier to have them identical (in many ways). However, I think its a bit of a waste to have staging replicated (for now, until maybe we get even more replication on the main one)... also worried they might ask why I need all that storage when I request it from ostack ;).

intarga:
> I agree, but I really would prefer to look into how to make it so I can start the notebook with different amounts of nodes. You are right it might be easier to have them identical (in many ways). However, I think its a bit of a waste to have staging replicated (for now, until maybe we get even more replication on the main one)... also worried they might ask why I need all that storage when I request it from ostack ;).

Another point in favour of two-node staging: having replication in staging, could potentially catch bugs related to replication that we might otherwise miss, like if a migration hangs or borks only in a replicated environment.

It’s not a big deal though, so do what you think is best 👌

ansible/roles/backups/tasks/main.yml
lcoram: Removed

ansible/roles/recovery/files/recovery_dump.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -e
+SECONDS=0
+
+echo "Starting restore of: $1 into $2"
+
+# we have 16 VCPUs on the VMs currently, so hopefuly -j 8 is ok 
+pg_restore -U postgres -d "$2" -j 8 "$1"
```
Lun4m, line 8: Probably should add `--exit-on-error` and `--clean` (to avoid "already exists" errors)? We should double check which exit codes are returned.

intarga (approved):

ansible/roles/recovery/files/recovery_dump.sh
Lun4m: But probably that's not a problem if creating the database from scratch

Lun4m:
Also the restore fails (when restoring the whole DB) if replication is on

PR #98, opened 2025-04-08
DROPS basic

jo-asplin-met-no:
DROPS is a sketch of how we might want to organize and think about products. Feel free to modify the ideas.


intarga:
@jo-asplin-met-no Am I correct in understanding that this is mostly meant as a reference to come back to later?

In that case maybe it would be best to leave this as an unmerged draft PR? That way it's still accessible as a reference, but doesn't add to our compile time, and lets it be easier to distinguish what is and isn't fully baked

PR #99, opened 2025-04-08
Optimise kvkafka ingestion

intarga:
## This PR includes:
- Moving XML message parsing to a sync thread to avoid choking the tokio executor, as it blocks longer than is recommended
- Batching messages for DB insertion so they can be done in parallel and take advantage of query pipelining without losing ordering that we need to ensure offset commits to the broker are correct
- Inserting in a transaction to avoid COMMIT on every INSERT
- Pipelining of label queries on the happy path where the labels already exist
- Using the `kvalobs` source-specific label for data ingested through kvkafka
- Parsing and ingestion of `original`, as I noticed it was missing (I'm guessing because we were originally only intending to use kvkafka for QC info?)
- Rework of the kafka e2e test, giving it much wider coverage of the module, and better reflecting how it runs in prod

## Things to note:
- In testing on the production VMs it was able to ingest 500-600 messages per second, which equates to it ingesting the queue ~250x faster than it is produced. It gets through a day's worth of messages in ~7 minutes, and the whole checked queue (2.6 million messages, ~26 million observations) in 1hr 20mins. This will probably vary based on how full the current partition is.
- While there is nitpick optimisation potential in terms of memcpy that can be avoided (allocating vectors, cloning structs), but as IO (waiting on postgres to complete INSERTs) is the main bottleneck now, I think we've reached the optimum effort to results ratio.


ingestion/src/kvkafka.rs
```diff
@@ -179,15 +183,128 @@ fn parse_message(message: &str) -> Result<Vec<RawDatum>, Error> {
...
 
+async fn create_timeseries(
+    conn: &mut PooledPgConn<'_>,
+    raw_datum: &RawDatum,
+    permit: Option<i32>,
+) -> Result<i64, Error> {
+    let transaction = conn.transaction().await?;
+
+    // TODO: should we re-check for an existing met label, since the first check was outside the
+    // transaction?
```
lcoram, line 194: Hmm, since hypothetically a label could have been created in the meantime? There is definetely a potential race condition here... and it might be important since often what happens is a bunch of new data starts coming in at once (say several elements for one station) and several timeseries have to be created. The worst case is that someone starts trying to dump a new timeseries, but with a bunch of old data... so each piece of data needs the timeseries to exist.  We almost need a lock of some sort on creation of timeseries?

ingestion/src/kvkafka.rs
```diff
@@ -179,15 +183,128 @@ fn parse_message(message: &str) -> Result<Vec<RawDatum>, Error> {
...
+
+    // TODO: currently we create a timeseries with null location
+    // In the future the location column should be moved to the timeseries metadata table
```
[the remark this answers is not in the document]
intarga: I didn't actually create this comment, I just moved it. It seems to date back to [this commit](https://github.com/metno/lard/commit/1472cb345a71bf3f47877797abcbe0f8d7215fa5) by @Lun4m. I'm not sure exactly what was meant by "timeseries metadata table"

I would take this comment as just a reminder to do "something" about location

ingestion/src/kvkafka.rs
lcoram: Ok, good to explain that it provides the lock. I had not understood that, but I need to go more deeply into the code... so far just a superficial pass.

intarga: This is fixed now, with a comment explaining the rationale behind the lock

ingestion/src/kvkafka.rs
```diff
@@ -94,6 +110,9 @@ fn create_consumer(brokers: &str, group_id: &str, topic: &str) -> LoggingConsume
         .set("auto.commit.interval.ms", "5000")
         // but only commit the offsets explicitly stored via `consumer.store_offset`.
         .set("enable.auto.offset.store", "false")
+        // if we don't have a starting offset, or it's out of range, start from the earliest
+        // available on the cluster
+        .set("auto.offset.reset", "earliest")
```
lcoram, line 115: Is there a way to explicitly set the offset? We had this for reingesting before, but it tried to "eat both ends" at once to get the new data while catching up with the old data...

ingestion/src/kvkafka.rs
```diff
@@ -179,54 +193,112 @@ fn parse_message(message: &str) -> Result<Vec<RawDatum>, Error> {
...
+                &raw_datum.kvid.station,
+                &raw_datum.kvid.paramid,
+                &raw_datum.kvid.typeid,
+                &raw_datum.kvid.level,
+                &raw_datum.kvid.sensor,
+            ],
+        )
+        .await?;
+
+    transaction.commit().await?;
```
lcoram, line 284: Do you need to explicitly unlock? or is it implicit after the transaction is commited / completed?

ingestion/src/kvkafka.rs
intarga: The broker stores the offset based on your consumer group, which we set as "lard" in the systemd unit. This option only applies if the broker doesn't have a saved offset for that group or it's not in range. If we want to reingest, we change the consumer group name, which is the same as what we did in ODA

ingestion/src/kvkafka.rs
intarga: Yes, locks only apply to a single transaction, and release when it completes. Postgres doesn't provide any way to explicitly release a lock other than committing or rolling back the transaction holding them.

ingestion/src/kvkafka.rs
intarga: Note: the default for this option is "latest" (well actually "largest", which is a synonym for "latest"), which will start reading from the front of the queue, which we definitely don't want.

Documentation for these config options can be found here:
https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md

ingestion/src/kvkafka.rs
lcoram: Ok, maybe just put a comment about how this releases the lock (for posterity / future people's understanding :)

intarga: Done

lcoram (approved):

lcoram:
I think @Lun4m  is also looking at this, so maybe wait for his review as well?

integration_tests/tests/end_to_end.rs
```diff
@@ -613,20 +639,125 @@ async fn test_kafka() {
...
-            .query_one("SELECT * FROM legacy.data", &[])
-            .await
-            .is_ok());
+        let open_conn = db_pools.open.get().await.unwrap();
+
+        // As we have no way to sync with message processing in kvkafka ingestion, we just keep
+        // trying to fetch data with a timeout
+        let timeout = std::time::Duration::from_secs(10);
+        let timeout_start = Instant::now();
+        loop {
```
Lun4m, line 656: Are we expecting for it to take 10 seconds to ingest a single message?? I honestly would just drop to 1 second and remove the loop, because if it takes longer there's definitely something wrong with the ingestion :sweat_smile: (I know it might be flaky, but still...)

Lun4m (changes requested):

ingestion/src/kvkafka.rs
```diff
@@ -299,76 +400,158 @@ async fn insert_kvdata(conn: &mut PooledPgConn<'_>, data: Vec<Datum>) -> Result<
...
 
     Ok(())
 }
 
-pub async fn process_message(
+async fn insert_batch(
     open_conn: &mut PooledPgConn<'_>,
     restricted_conn: &mut PooledPgConn<'_>,
-    message: &str,
+    raw_data: &mut [(Vec<RawDatum>, (i32, i64))],
```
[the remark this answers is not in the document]
intarga: Removed

ingestion/src/kvkafka.rs
```diff
@@ -179,54 +193,112 @@ fn parse_message(message: &str) -> Result<Vec<RawDatum>, Error> {
...
+            "INSERT INTO labels.met \
+        (timeseries, station_id, param_id, type_id, lvl, sensor) \
+    VALUES ($1, $2, $3, $4, $5, $6)",
+            &[
+                &timeseries_id,
+                &raw_datum.kvid.station,
+                &raw_datum.kvid.paramid,
+                &raw_datum.kvid.typeid,
+                &raw_datum.kvid.level,
+                &raw_datum.kvid.sensor,
```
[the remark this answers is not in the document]
intarga: Hmm, I would prefer to just leave it as an issue, since we already have an issue, and it requires changes in a lot more places than this

ingestion/src/kvkafka.rs
```diff
@@ -10,13 +13,26 @@ use serde::Deserialize;
...
+// Query to get a tsid from the relevant source-specific label
+const QUERY_GET_MET_STR: &str = r#"
+    SELECT timeseries FROM labels.kvalobs
+        WHERE station_id = $1
+        AND param_id = $2
+        AND type_id = $3
+        AND (($4::int IS NULL AND lvl IS NULL) OR (lvl = $4))
+        AND (($5::int IS NULL AND sensor IS NULL) OR (sensor = $5))
+    "#;
+
```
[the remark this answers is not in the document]
Lun4m: No no, we can do it later :)

integration_tests/tests/end_to_end.rs
Lun4m: Weird, on my machine it "consistently" takes 0.3 s :thinking: maybe it's cached?

ingestion/src/kvkafka.rs
```diff
@@ -235,62 +313,85 @@ async fn filter_and_label_kvdata(
...
+    for i in fails {
+        let tsid = create_timeseries(conn, &raw[i].0, raw[i].1).await?;
         data.push(Datum {
             tsid,
-            obstime: raw_datum.obstime,
-            kvdata: raw_datum.kvdata,
+            obstime: raw[i].0.obstime,
+            kvdata: raw[i].0.kvdata.clone(),
         });
     }
```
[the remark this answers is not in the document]
Lun4m: Yes, that might be true! I'm not sure I 100% understand what you mean by prioritizing the happy path here, aren't we processing both paths in the same function/thread? 

Anyway I find this version a bit more readable than using enumerate and tuple indexing, but I guess it's simply preference.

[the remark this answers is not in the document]
intarga: Happy path is the case where fails is empty, which should be basically always

ingestion/src/kvkafka.rs
```diff
@@ -179,54 +193,112 @@ fn parse_message(message: &str) -> Result<Vec<RawDatum>, Error> {
...
+
+    transaction.commit().await?;
+
+    Ok(timeseries_id)
+}
 
-        // let transaction = conn.transaction().await?;
+async fn label_kvdata(
+    conn: &mut PooledPgConn<'_>,
+    raw: Vec<(RawDatum, Option<i32>)>,
```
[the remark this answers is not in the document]
intarga: Done. Changed this to raw_data, the others to raw_buffer, and imported the `PermitId` type alias to make the type clearer

ingestion/src/kvkafka.rs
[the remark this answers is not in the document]
Lun4m: Yeah, I understand that, I just didn't see the prioritization part, but I guess we are saving an iteration over `raw_data` with your original version.

ingestion/src/kvkafka.rs
```diff
@@ -299,76 +400,158 @@ async fn insert_kvdata(conn: &mut PooledPgConn<'_>, data: Vec<Datum>) -> Result<
...
     // for this? The kafka group coordinator should automatically assign partitions to consumers
     // such that the group covers all partitions, and we shouldn't have to worry about it. On that
     // note though, should be spawn a consumer task for each partition? It should increase our
     // throughput
-    const TOPIC: &str = "kvalobs.production.checked";
-    let consumer = create_consumer(BROKERS, &group, TOPIC);
+    let consumer = create_consumer(brokers, group, topic);
+
+    // Channel buffer size here is based on pure vibes, feel free to change it
+    let (parse_tx, mut parse_rx) = tokio::sync::mpsc::channel::<(String, (i32, i64))>(1);
```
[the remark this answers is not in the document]
intarga: Done

integration_tests/tests/end_to_end.rs
intarga: ?

Lun4m: I had commented out the `loop` (and added a `tokio::time::sleep` instead of the timeout), but without changing the `if let Ok(data_row) = ...` :sweat_smile:

Lun4m (approved):

PR #100, opened 2025-04-10
More ansible/migration fixes

Lun4m:
- Fixed some small issues with the migration code, so multiple migration commands can be piped safely.
- Converted existing migration playbooks used only to import the database schemas to instead set up migrations from kdvh and kvalobs.


ansible/roles/pg/tasks/configure/create_standby.yml
```diff
@@ -27,8 +27,9 @@
   become: true
 
 # https://www.repmgr.org/docs/current/quickstart-standby-clone.html
+# TODO: maybe force '-F' should not be used lightly
```
intarga, line 30: Can you add some more context here?

ansible/README.md
```diff
@@ -310,3 +310,10 @@ ansible-inventory -i inventory.yml --graph
 
 ansible servers -m ping -u ubuntu -i inventory.yml
 ```
+
+#### Encoding single variables
+
+```terminal
+# Notice the leading space
+ ansible-vault encrypt_string '000.000.000.000' --name ansible_host >> host_vars/server.yml
+```
```
intarga, line 319: Interesting, I just use `ansible-vault edit` for everything

intarga: Oh I see. I personally prefer only file-level encryption, but I don't mind doing it this way either. Thanks for documenting it in any case.

ansible/roles/pg/tasks/configure/create_standby.yml
Lun4m: Actually, I think it's fine using force both here and for standby register, since this `yml` file is only called when we first set up the VMs or when we need to rebuild a replica after the migrations. So it makes sense simply removing the TODO

Lun4m: Removed the TODO and FIXME

ansible/README.md
Lun4m: Changed the directory structure of `host_vars`

PR #101, opened 2025-04-23
Add IDF station endpoint

Lun4m:
Some comments:

1. This mostly implements the same ODA output schema, the only difference being in the availability output, where ODA returns

   ```json
   "stations": [
       {
           "available": {
               "from": "2013-12-31T00:00:00Z",
               "to": "2024-12-31T00:00:00Z"
           },
           "extra": {
               "firstyearofperiod": "",
               "lastyearofperiod": "",
               "numberofseasons": "",
               "qualityclass": "",
           },
           "id": {
               "stationid": 18205
           }
       },
   ]
   ```

   while we currently return all the metadata

   ```json
   "stations" : [
       {
           "stationId": 18205,
           "firstYearOfPeriod": 2013,
           "lastYearOfPeriod": 2024,
           "numberOfSeasons": "not sure what this actually is",
           "qualityClass": 3,
           "seedParameter": 0,
           "updatedAt": "2025-01-01",
       }, 
   ]
   ```

1. I haven't implemented the ingestion of IDF data, which as far as I understand comes as a CSV file. Should we have a POST endpoint in the `ingestor` (maybe we should rename it `ingress`?)? Or should it be in the `egress` crate?
   EDIT: from the discussion below, a separate client should be developed.

1. After speaking with Ketil, this endpoint is currently only used by Seklima. It doesn't seem worth it to support selecting multiple station/duration/frequency combinations.


db/006-reports.sql
```diff
@@ -0,0 +1,26 @@
+CREATE SCHEMA IF NOT EXISTS reports;
+
+CREATE TABLE IF NOT EXISTS reports.idf_station_timeseries (
```
intarga, line 3: Would be nice to point to the struct documentation here, in case someone reaches this first and is confused or assumes there is no documentation.

egress/src/reports/idf_station.rs
```diff
@@ -0,0 +1,238 @@
...
+    tsid: i32,
+    number_of_seasons: i32,
+    // TODO: should we have these instead?
+    // fromtime: Option<DateTime<Utc>>,
+    // totime: Option<DateTime<Utc>>,
+    first_year_of_period: i32,
+    last_year_of_period: i32,
+    quality_class: i32,
+    seed_parameter: i32,
+    updated_at: chrono::DateTime<Utc>,
```
intarga, line 55: Docstrings for the fields would be appreciated, it's not obvious what `seed_parameter` and `number_of_seasons` are for example.

egress/src/reports/idf_station.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct IdfStationInfo {
```
intarga, line 60: This struct seems to be subset of `IdfMetadata`. It's unclear why they're separate? I assume there's a good reason, so we should clarify this with a comment and/or clearer naming and/or module structure

egress/src/lib.rs
```diff
@@ -131,6 +133,11 @@ pub async fn run(pool: PgConnectionPool, cancel_token: CancellationToken) {
             get(timeslice_handler),
         )
         .route("/latest", get(latest_handler))
+        .route(
+            "/reports/idf/station",
+            get(idf_station_availability_handler),
+        )
+        .route("/reports/idf/station/:station_id", get(idf_station_handler))
```
intarga, line 140: I think nested routers would make this neater, and would let us move the routing for idf into the idf module.

integration_tests/tests/idf_station.rs
```diff
@@ -0,0 +1,173 @@
...
+        for val in resp.values {
+            // Since the values get converted from 'mm' to 'lsha'
+            // we should not find them in the response
+            assert!(!data.intensity.contains(&val.intensity));
+            assert_ne!(val.lower_interval, data.lower);
+            assert_ne!(val.upper_interval, data.upper);
+        }
+    })
+    .await
+}
```
intarga, line 173: Would be good to check that the data is correct not just that it's there

integration_tests/tests/common/mod.rs
intarga, line 1: Good initiative modularising this! I'm going to have a nasty conflict though, because I've heavily reworked the kafka tests 😅

intarga (changes requested):

egress/src/reports/idf_station.rs
Lun4m: True, this is only used for the 'availability' response. The intention was to mimic what ODA is doing, but since I changed that (point 1. in the first comment), I guess we can return all the metadata instead.

egress/src/reports/idf_station.rs
Lun4m: I agree, they are parameters used for generating the values and I still have to figure out what they mean myself (I haven't read the whole document/code)

egress/src/lib.rs
Lun4m: Good point!

integration_tests/tests/common/mod.rs
Lun4m: Let me take a look at your PR

intarga: It's not a problem! I'm just groaning at my future work 😛

egress/src/lib.rs
Lun4m: Done!

db/006-reports.sql
Lun4m: Done!

egress/src/reports/idf_station.rs
Lun4m: Done! But I'll wait for Ketil to confirm the description makes sense (maybe we should double check with Lars/Julia?)

egress/src/reports/idf_station.rs
intarga: Yes, thank you

integration_tests/tests/idf_station.rs
Lun4m: Done, not sure it's the ideal way though

egress/src/reports/idf_station.rs
ketilt: Lars/Julia would be the authorities on documentation of IDF yes

db/006-reports.sql
ketilt, line 1: Do we want to define and lock the format in the DB itself? IDF is one of many bespoke data formats that we might want to support. My guess was that S3 was a better way of storing them, allowing us to more simply add new report types without needing to make fundamental changes to the system. In ODA I tried to facilitate this further by having the report schemas be handled by the API, allowing users to define them without needing developer access or redeploying any systems. The format of the report may also change as the method is updated. This was a hassle in KDVH where the original table format for IDF is outdated, and as a result frost v0 still gives out years old data.

For IDF and some other reports we'll also want to have versioning, so that you can request the report active at a given date. In ODA, I added the optional field `operatingPeriods` because older IDF reports include this. The ones listed here with NOT NULL should keep that, as they are considered required in the OutputSchema:
https://viewer.oda.met.no/productfunctions/649282252448399361

intarga: This actually raises a deeper question: how "in the loop" should we be on this? I'm inclined to think that either we should own the whole process including generating the reports (this was what I assumed to be the case before Manuel started working on this), or lard should not be involved at all. Both approaches have advantages:

- If we own the whole process, we don't have a third-party dependency, and can guarantee stability and integrity
- If we are completely outside the loop, Julia and Lars can iterate and modify the format without having to go through us

The current state seems to be the worst of both worlds. We should probably consult Julia and Lars on which they would prefer. If their reason for involving us is that they're uncomfortable maintaining the infrastructure themselves, we can offer help with that without directly involving lard.

Lun4m: S3 could definitely be an option, we would need to define the file format (one CSV per station? Metadata in the header row? Policy on update? Are we the ones supposed to prepare these files or is it on the ones that generate the IDFs?).
Anyway I think we would still need to change stuff on our side if the metadata/method changes (unless the metadata header in the files is simply treated as a generic map, which is not ideal?).

Regarding `operatingPeriods`, looking at the ODA code I thought it was not used :sweat_smile: 
But doesn't the endpoint only return the latest IDF values anyway?

ketilt: @intarga How you delineate "the whole process" in this case? IDF and most other reports can be considered as research results. From a system perspective I would see them as standardized deliveries from an external producer, much like our observational data.

So it does make sense to lock the structure. This is how ODA operated de facto, since only I created and modified report definitions in the API. But I could do so without having to redeploy ODA. The schema defined would validate all reports going in, so no garbage can enter the system. But definitely I would prefer not having to be that gatekeeper, and ideally I would have set up a GUI where Julia and Lars could upload a CSV without my involvement. Would that count as outside the loop?

I'd still be needed for updates to the format. But that is also the case if this is defined in the SQL schema itself.

@Lun4m We've kept it all in one CSV until now, but that can be changed. The only advantage I can see of a large CSV is if this is to be entered into a web ui, and one file would be easier. I would expect the producers to prepare the file in the right format from the start. We're almost there for IDF already, we just don't have a documentation reference for what the format should be.

Yes, currently ODA does not support versioning of IDF. So for decommissioning ODA, this is not necessary.

intarga: @ketilt If these reports are going to be generated outside our system and don't have a *direct* connection to the observational data, I would say this doesn't belong in lard/ODA and should have its own standalone service.

Whether it's in lard or its own reports service, the discussed concerns still exist. The system you describe offers only the barest integrity guarantees:
- No checks against anything not in the report itself (i.e. that the station exists and was reporting data for the whole window the report purports to be relevant to)
- No assurances that reports actually get generated, or follow a particular schedule
- No mechanism for existing entries to be migrated or regenerated on a schema change
- No assurances of code-review and testing of changes to the generation code

In my view, whether these concerns matter should be decided by how the reports are used. 

If the reports are operationally critical (i.e I believe IDFs are used to plan engineering projects) then these concerns matter and make a good argument to formalise them (personally I'm not happy to tell people my system delivers operationally critical reports without these guarantees). Researchers can still experiment with modifications outside our system, but for changes to hit end users they must go through us.

If the reports aren't critical, and are just for researchers to tinker with on their own terms, then I wouldn't even bother with schema validation to be honest.

Lun4m:
Ok, I'll drop the database tables in favor of reading CSVs files from object store.

ketilt:
> Ok, I'll drop the database tables in favor of reading CSVs files from object store.

Although the input CSV is fine in terms of data sanity and algorithm integrity, the file has been manually created and I've had to request formatting fixes each year. There've been semicolon delimiters, quoted values and missing columns, mostly hailing from Microsoft Excel. So some sort of validation is important.

I don't want to create more work for you guys, so if the database tables functions as a validation already, then that's probably fine. The flexibility argument still applies, but not at the cost of validation and control over the format of what is used operationally.

Lun4m:
Mmm, ideally the CSV validation should be responsibility of whoever generates the file, however my idea was to parse and split it by stations on upload, so we can probably catch formatting errors there.

ketilt:
> Mmm, ideally CSV validation should be responsibility of whoever generates the files (?), however my idea was to parse and split it by stations on upload, so we can probably catch formatting errors there.

I agree that it is the responsibility of the producer, nobody else should need to fix their inputs for them (as I have), but somewhere we definitely need to have a security catch. My vision is a web GUI where producers can insert their CSVs for validation and feedback, with documentation and specs close by. One scenario is that this GUI is not part of LARD and would be made as a standalone application by me or someone else. The resulting files would be saved on S3 to be picked up by Frost or LARD, and the integrity of the reports is then ensured in the agreement between this GUI service and LARD.

The JSON schema validation I implemented in ODA was an unfinished solution towards this vision. I'll leave if to you guys to decide how you want to scope this. If you go for no dedicated validation in LARD, then do keep me in the loop so that I can help ensure the data integrity until a more robust validation is in place somewhere.

intarga:
@ketilt 

> It's probably the most scrutinized product that we have. (EDIT: Scrutinized in terms of the data, not the CSV formatting!)

These two things don't comport IMO. Of the five guarantees we should have (the four I listed, plus schema compliance for new data), there's only one we can verify right now, and the fact that it routinely fails does not bode well for the other four.

From a practical perspective, it doesn't make sense to me that the data can be so scrutinised without the formatting being scrutinised too. Either:
- The data is scrutinised before it is manually compiled and exported - in which case it's susceptible to the same human error as the formatting (i.e mixing up two files, misplacing a row or column)
- The data is scrutinised after compilation and export - in which case, why does this scrutiny not catch obvious formatting mistakes?
- The data is scrutinised after submission to us - in which case it's too late

ketilt:
There are some good points here that we could put in a risk analysis. Does it make sense to have a separate one for IDF station data? I could set one up if you want, and we could use that to communicate and decide how to alleviate these concerns together with the IDF producers. We could also use it to distribute responsibility and accountability if that is a concern.

What I believe we can achieve, is:
* The data is always autogenerated, so no manual inteference before exporting. (Any manual QC should be either assertion only, or grounds for changing parameters until the produced file is OK.)
* File formatting should not be altered, and is checked manually or automatically to be up to spec.

The validation I did in ODA was an attempt at that last point. Passing it is not a green light for everything the data may represent, it's only a sanity check that the system will not break trying to distribute it. So I've been reluctant to see such a check go. If it's a better division of responsibility to require full data integrity from the producer, then that's worth a shot.

Lun4m:
Ok, I finished the S3 implementation and refactored a bit the integration tests. I will send a message to Julia tomorrow regarding the missing documentation and then mark this one ready for review!

Lun4m:
Well, I guess this is actually ready for review, even if the doc comments might be updated later, before we merge it,

egress/src/reports/idf_station.rs
```diff
@@ -0,0 +1,375 @@
...
+    }
+}
+
+/// Metadata and parameters used for fitting IDF values
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct IdfMetadata {
+    /// MET station identifier
+    station_id: i32,
+    /// Number of three month periods considered in the calculation
```
ketilt, line 68: (Unless those who know better have said otherwise) I interpret "season" in the IDF sense as referring to the portion of the year that had a continual period with no frost. Similar to season in the term "growing season". If so, it's numerically equivalent to the number of years of data. But years would be incorrect since it's not whole years

egress/src/reports/idf_station.rs
```diff
@@ -0,0 +1,375 @@
...
+    number_of_seasons: i32,
+    // TODO: should we have these instead?
```
ketilt, line 70: I think dates could give an impression of temporal continuity, when in reality only portions (of variable length) of data from each year have been used

egress/src/reports/idf_station.rs
```diff
@@ -0,0 +1,375 @@
...
+    // fromtime: Option<DateTime<Utc>>,
+    // totime: Option<DateTime<Utc>>,
+    /// First year considered in the precipitation timeseries
+    first_year_of_period: i32,
+    /// Last year considered in the precipitation timeseries
+    last_year_of_period: i32,
+    /// Quality of the timeseries used for the calculation
+    // TODO: weighs length, resolution, and? Is there a proper definition?
```
ketilt, line 78: Lars Grinde might be able to provide a better description. I do not know it. It is conceivable that the sensor performance and exposure could have an effect, as well as data coverage over the full time period. Also outdated series from discontinued stations might be demoted over time? But these are only vague guesses, and I think the definition is more concrete.

egress/src/reports/idf_station.rs
```diff
@@ -0,0 +1,375 @@
...
+fn mm_to_lsha(val: f64, duration: i32) -> f64 {
+    1e4 / 60.0 * val / duration as f64
+}
+
+// TODO: need blocking thread?
+fn parse_values_csv(bytes: &[u8], unit: IdfUnit) -> Result<(IdfMetadata, Vec<IdfValue>), Error> {
+    // flexible allows us to store metadata in the header
+    let mut reader = csv::ReaderBuilder::new().flexible(true).from_reader(bytes);
+
+    // TODO: duplicated metadata record in station csv header row, are there better options?
```
ketilt, line 144: Was there duplication in the header in the example file, or is this a contingency check?

ketilt: That makes sense, one file per user request basically, the first one just browsing the metadata before requesting station IDFs

ketilt:
As a matter of curiosity, would be it challenging to implement access restrictions for reports like the IDF station?

A hazard warning meteorologist colleague has been requesting additional IDF Station data be made available in Frost for use on Geoweb. Lars Grinde produces these data as hoc. They have the same format as the other IDF Station data, but they are based on short timeseries, are highly uncertain and should not be used for dimensioning purposes like the other data.

ODA never implemented restricted reports and I have not been keen on the idea, but I thought I should mention it.

ketilt:
> > As a matter of curiosity, would be it challenging to implement access restrictions for reports like the IDF station?
> 
> I don't think it's going to be any more difficult than implementing it for other restricted access resources. However off the top of my head the case you are referring to doesn't seem to fall into any of the current permits? So we would probably need a new permit and to include it in the metadata?

That is correct. There are other questions to be resolved too, and I wouldn't spend time on it now. Knowing it's technically possible is good to know for long term planning

egress/src/reports/idf_station.rs
Lun4m: Yes, it's actually just the number of years. I also added the reason why a year could be excluded :+1:

egress/src/reports/idf_station.rs
Lun4m: I guess the next step would be to modify the IDF importer from ODA to perform the splitting of the input CSV? I'm not sure we will have much time to develop a web UI before the beta, but that could definitely come at a later time. What do you think?

.env.test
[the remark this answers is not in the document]
Lun4m: Done!

compose.yml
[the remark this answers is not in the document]
Lun4m: Done!

justfile
```diff
@@ -1,59 +1,57 @@
...
-    @ if docker ps -a | grep lard_tests > /dev/null; then just _clean > /dev/null; fi
+psql db="lard":
+    @ docker exec -it lard_postgres psql -U postgres -d {{db}}
+
+_setup: _clean
+    docker compose -f compose.yml up -d
+    @ echo "Setting up S3 bucket..."
+    @ python3 -m venv $LARD_VENV
+    @ $LARD_VENV/bin/python3 -m pip install awscli-local[ver1] > /dev/null
+    @ $LARD_VENV/bin/awslocal s3 mb s3://$S3_BUCKET_NAME > /dev/null
```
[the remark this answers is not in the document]
Lun4m: Unfortunately there seem to be a bug in `rust-s3` when you try to create a bucket in `localstack`. Added a TODO comment with a bit of context and with the aim of removing this code when that bug is resolved.

egress/src/reports/idf_station.rs
ketilt: Yes, no point spending time on a GUI at this point I think. I had a brief chat with Lars and Julia today about standardizing the CSV output and I think that is definitely manageable

db/006-reports.sql
Lun4m: I'm marking this conversation as resolved, but feel free to create an issue about standards for external contributions/products that should be available in LARD.

Lun4m:
I guess this can be approved and merged now? For the "ingestion" of these IDF files we will simply adapt the scripts we already have in ODA (https://github.com/metno/lard/pull/101#discussion_r2098308282).

intarga (approved):

PR #112, opened 2025-05-08
Levels from stinfosys to convert from default 0

lcoram:


ingestion/src/levels.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+pub fn param_get_level(
+    level_table: Arc<RwLock<ParamLevelTable>>,
+    param_id: i32,
+) -> Result<Option<i32>, Error> {
+    let level_table = level_table.read().map_err(|e| Error::Lock(e.to_string()))?;
+
+    if let Some(param_level) = level_table.get(&param_id) {
+        // Convert level to cm
```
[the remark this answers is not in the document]
lcoram: And this makes me realize actually we have to convert all the levels, not just the ones that are 0... So they are in cm. But can we assume they are in meters or how do we know their scale?

ingestion/src/levels.rs
```diff
@@ -0,0 +1,102 @@
...
+
+pub fn param_get_level(
+    level_table: Arc<RwLock<ParamLevelTable>>,
+    param_id: i32,
+    level: i32,
+) -> Result<Option<i32>, Error> {
+    let level_table = level_table.read().map_err(|e| Error::Lock(e.to_string()))?;
+
+    if let Some(param_level) = level_table.get(&param_id) {
+        // if level is 0, replace with default
```
lcoram, line 86: Convert to default

ingestion/src/levels.rs
```diff
@@ -0,0 +1,102 @@
...
+        let mut lvl = level;
+        if level == 0 {
+            lvl = param_level.hlevel; // this is the default
+        }
+        // Convert level to cm (use scale from stinfosys)
+        // scale for things that are currently in m is 0, so need to shift
```
lcoram, line 92: Use scale to convert to cm

ingestion/src/levels.rs
```diff
@@ -0,0 +1,102 @@
...
+        let scale = param_level.hlevel_scale;
+        return Ok(Some(match scale {
+            0 => lvl * 100, // convert from meters
+            -2 => lvl,      // already in cm
+            // oh dear, this isn't meters or cm?
+            _ => unreachable!(), // Shouldn't happen due to check when reading from stinfosys!
+        }));
```
lcoram, line 99: We are only looking for the 2 scales that currently exist in stinfosys

ingestion/src/levels.rs
```diff
@@ -0,0 +1,102 @@
...
+            &[],
+        )
+        .await?;
+
+    // build hashmap of param permits
+    let mut param_level = HashMap::new();
+
+    for row in rows {
+        let hlevel_scale: i32 = row.get(1);
+        if ![0, -2].contains(&hlevel_scale) {
```
lcoram, line 64: but do we have stuff that has a hlevel but not a scale? @pmlefeuvre-met

lcoram:
So the migration part of the task is being dealt with in another branch (and in golang):
https://github.com/metno/lard/commit/ae52f6cde3aaf3f799cd1bbcb99d324149054a01

ingestion/src/kldata.rs
```diff
@@ -395,7 +405,7 @@ pub async fn filter_and_label_kldata<'a>(
                                 &chunk.station_id,
                                 &param.id,
                                 &chunk.type_id,
-                                &lvl,
+                                &level,
                                 &sensor,
```
lcoram, line 409: @intarga this handles the obsinn ingestion part?

Lun4m: Yes!

ingestion/src/levels.rs
```diff
@@ -0,0 +1,102 @@
...
+            // currently only have 0 and -2, aka m and cm
+            param_level.insert(
+                row.get(3),
+                Level {
+                    hlevel: row.get(0),
+                    hlevel_scale,
+                },
+            );
+        }
```
Lun4m, line 87: I would remove this if condition and add a runtime warning/error if the level is not 0 or -2 instead (in `param_get_level`? Otherwise I we should log here the params with weird scales?

```suggestion
        param_level.insert(
            row.get(2),
            Level {
                hlevel: row.get(0),
                hlevel_scale: row.get(1),
            },
        );
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    // build hashmap of param permits
+    let mut param_level = HashMap::new();
+
+    for row in rows {
+        let hlevel_scale: i32 = row.get(1);
+        if ![0, -2].contains(&hlevel_scale) {
+            // currently only have 0 and -2, aka m and cm
+            param_level.insert(
+                row.get(3),
```
Lun4m, line 81:
```suggestion
                row.get(2),
```

lcoram:
Unlinked issue 110 I think, and have made a first pass at a comment for the levels.

ingestion/src/levels.rs
```diff
@@ -0,0 +1,116 @@
...
+// Both in obsinn and in kafka the timeseries label includes a level.
+// Unfortunately (for historical reasons) 0 is used to mean 'default', and does
+// not actually always mean that it has 0 height. We will keep the label as it
+// comes in for the obsin, kvalobs, and kdvh labels (to preserve provenance).
+//
+// In Lard for the MET labels we wished to no longer have 0 be default, but
+// rather replace it with the actual parameter's default height. Additionally,
+// the scale of level can differ, so we chose to standardize it to cm.
+// These conversions are handled in this file, and currently rely on the
+// param table in stinfosys.
```
lcoram, line 12: Comment ok?

Lun4m (approved):
Looks good to me!

ingestion/src/levels.rs
intarga:
```suggestion
//! Level as used here refers to meteorological height above ground (or depth
//! if negative) of a sensor in centimetres.
//!
//! "Meteorological" here means the number may not be the sensor's exact
//! physical height, but a standard height closeby that this has been
//! approved for comparison to other stations based on. Also, for certain
//! parameters, this height is not taken relative to the ground (i.e for wind
//! it is taken relative to the top of any uneveness in the nearby landscape).
//! Notably this differs from but is easily confused with:
//! - elevation/hamsl - typically refers to the vertical distance between mean
//!   mean sea level and the ground the station stands on.
//! - altitude - typically refers to the vertical distance between mean sea
//!   level and the sensor.
//!
//! This differs from previous level semantics at Met (notably in ODA, kvalobs,
//! stinfosys), where level typically meant metres (though sometimes cm
//! depending on the parameter), cannot be 0 as 0 is a special value
//! indicating that the level is a default for the param (listed in stinfosys),
//! and NULL is always equivalent to 0.
//!
//! Justifications for the old semantics (incomplete and inferred, as the
//! people responsible are no longer with us):
//! - `0 = default` reduces configuration work
//! - `NULL = 0` means default levels can be omitted from messages reporting
//!   reporting data from stations, saving some bytes, which used to be quite
//!   expensive.
//!
//! Justifications for changing to the new semantics:
//! - `0 = default` encourages mistakes. A pattern observed at met is people
//!   setting level = 0 for a new sensor to get it working/reporting quickly,
//!   assuming someone will fix it later, which doesn't happen.
//! - `0 = default` makes it impossible to represent a level that is genuinely
//!   0, which does happen.
//! - `0 = default` is not what most end users expect meaning we either
//!   increase the burden on them, or convert to a different scheme at request
//!   time, which is redundant and confusing.
//! - `0 = default` is a friction point with our international collaborators
//!   who find it arcane (notable exception: the UK met office)
//! - `NULL = 0` removes the ability to use NULL for cases where we don't know
//!   the level, or it isn't relevant.
//! - Inconsistent units for level are likely a result of the system originally
//!   being designed for integer metres only, and then having to tack on sub-
//!   metre levels. Might as well fix that while we're making changes.
//!
//! These semantics do not apply to levels in the `kvalobs` and `obsinn` labels
//! as those source-specific labels are meant to reflect how the data was
//! reported in to lard, which in those cases is with the old semantics.
```

`//!` is used because this is a [doc comment](https://doc.rust-lang.org/reference/comments.html#r-comments.doc). `//!` documents the file, while `///` documents the object under it.

My suggested changes are just suggestions, of course feel free to disagree and/or refine further

ingestion/src/levels.rs
```diff
@@ -0,0 +1,102 @@
...
+impl Level {
+    pub fn new(hlevel: i32, hlevel_scale: i32) -> Level {
+        Level {
+            hlevel,
+            hlevel_scale,
+        }
+    }
+}
+
+type ParamID = i32;
```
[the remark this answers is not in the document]
intarga: I'm not sure either. Maybe we should makes a types module in the `util` crate for things like this, since it's likely to be shared with egress and other utilities too?

Feel free to ignore it and I can pick it up in my PR

ingestion/src/levels.rs
intarga: Disagree! I think it makes more sense to catch these when we load from stinfosys, not when we handle a datum. With the change we will fail to ingest data that uses it, which I don't think is ideal, and we lose our chance to detect the problem before that happens.

I agree that we should `error!` log and maybe have a metric too though

intarga: Given we eventually won't have kvalobs to fall back on, I think it's better to set the precedent of ingesting something that needs fixing, rather than risk missing the data entirely. We will need to do corrections to data based on stinfosys changes and other mistakes anyway, which falls to the content managers

intarga: Seems reasonable to me to only check at load time, and exclude unsupported entries from the map.

That way the data will have lvl = `NULL` in `labels.met` and their original hlevel in `labels.kvalobs`/`labels.obsinn`.

This should be easy to find a fix, because affected timeseries will have `NULL` in `labels.met` and `NOT NULL` in their source specific label, that can only be caused by this. Not to mention we should know what is affected simply because we know what the new param is in stinfosys, and can just look for timeseries with that paramid.

ingestion/src/main.rs
```diff
@@ -39,6 +39,13 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     ));
     let background_permit_tables = permit_tables.clone();
 
+    // Levels tables handling (needs connection to stinfosys database)
+    let level_table = Arc::new(RwLock::new(
+        levels::fetch_levels(&stinfo_conn_string).await?,
+    ));
+    // maybe need this later?
+    let _background_level_table = level_table.clone();
```
intarga, line 47: I just noticed we aren't refreshing this in the background like we are the permit tables. We should probably do that.

db/002-labels.sql
```diff
@@ -7,6 +7,7 @@ CREATE TABLE IF NOT EXISTS labels.met (
     param_id INT4,
     -- TODO: Maybe change this as we reevaluate type_id's usefulness and future at met?
     type_id INT4,
+    -- NOTE: this level will be converted to always be centimeters and to replace default with the actual level value
```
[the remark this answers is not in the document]
lcoram: Taken in recent commit

ingestion/src/main.rs
lcoram: Nominally added in last commit, but not with generics and such as mentioned.

ingestion/src/levels.rs
lcoram: Changed back as I saw fit based on this conversation in recent commit

ingestion/src/levels.rs
```diff
@@ -77,14 +113,16 @@ pub async fn fetch_levels(stinfo_conn_string: &str) -> Result<ParamLevelTable, E
...
+        if ![0, -2].contains(&hlevel_scale) {
+            // currently only have 0 and -2, aka m and cm
+            param_level.insert(
+                row.get(2),
+                Level {
+                    hlevel: row.get(0),
+                    hlevel_scale: Some(hlevel_scale),
+                },
+            );
+        }
```
intarga, line 125: Since we mentioned wanting to log this:

```suggestion
        } else {
            error!("Invalid hlevel_scale found in stinfosys: {}", hlevel_scale);
        }
```

Probably also need
```
import tracing::error;
```
at the top

ingestion/src/main.rs
intarga: We can take that up another time! Thanks for leaving a TODO

intarga (approved):
@pmlefeuvre-met mentioned he has some further refinements to the doc comments

ingestion/src/levels.rs
lcoram: Thanks! Done

ingestion/src/levels.rs
```diff
@@ -0,0 +1,159 @@
...
+//! "Meteorological" in this case means the number may not be the sensor's
+//! exact physical height, but a standard height close-by which has been
+//! approved for comparison to other stations. Also, for certain
+//! parameters, this height is not taken relative to the ground (i.e for wind
+//! it is taken relative to the top of any uneveness in the nearby landscape).
+//! Notably this differs from but is easily confused with:
+//! - elevation/hamsl - typically refers to the vertical distance between mean
+//!   mean sea level and the ground the station stands on.
+//! - altitude - typically refers to the vertical distance between mean sea
+//!   level and the sensor.
```
pmlefeuvre-met, line 13:
```suggestion
//! "Meteorological" in this case means that the height may not be the sensor's
//! exact physical height, but a standard height close-by which has been
//! approved for comparison to other stations. Also, for certain
//! parameters, this height is not taken relative to the ground (i.e for wind
//! it is taken relative to the top of any uneveness in the nearby landscape).
//! Notably meteorological height differs from but is easily confused with:
//! - elevation/height above sea level (hamsl) - typically refers to the vertical
//!   distance between mean mean sea level and the Earth surface,
//!   in other words the ground the station stands on.
//! - altitude - typically refers to the vertical distance between mean sea level 
//!   and the sensor at any point in the atmosphere (ex.: flight, radiosonde,...).
//!  
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,159 @@
...
+//!
+//! This differs from previous level semantics at Met (notably in ODA, kvalobs,
+//! stinfosys), where level typically meant metres (though sometimes cm
+//! depending on the parameter). The level could not be 0 as 0 was used a special
+//! value indicating that the level is a default for the param (default value
+//! found in stinfosys), and NULL is always equivalent to 0 (when the data is meant
+//! to have a level, which not all data does).
```
pmlefeuvre-met, line 20:
```suggestion
//! This differs from previous level semantics at Met (notably in ODA, kvalobs,
//! stinfosys), where level typically meant metres (though sometimes centimetres
//! depending on the parameter and as defined by the value hlevel_scale). 
//! The level could not be 0 metre as 0 was used as a special value indicating that 
//!  the level is a default for the param (default value is defined in standard_hlevel
//!  as found in stinfosys), and NULL is always equivalent to 0 (when the data is 
//! meant to have a level, which not all data does).
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,159 @@
...
+//! - `NULL = 0` means default levels can be omitted from messages reporting
+//!   reporting data from stations, saving some bytes, which used to be quite
+//!   expensive.
+//!
+//! Justifications for changing to the new semantics:
+//! - `0 = default` encourages mistakes. A pattern observed at met is people
+//!   setting level = 0 for a new sensor to get it working/reporting quickly,
+//!   assuming someone will fix it later, which doesn't happen.
+//! - `0 = default` makes it impossible to represent a level that is genuinely
+//!   0, which does happen.
```
pmlefeuvre-met, line 34:
```suggestion
//!   0, which does happen (i.e. surface parameters such as surface temperature)
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,159 @@
...
+//! - `0 = default` is not what most end users expect meaning we either
+//!   increase the burden on them, or convert to a different scheme at request
+//!   time, which is redundant and confusing.
```
pmlefeuvre-met, line 37:
```suggestion
//! - `0 = default` is not what most end users want (although they got used to it) 
//!    meaning we either increase the burden on them, or convert to a different
//!    scheme at request time, which is redundant and confusing.
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,159 @@
...
+//! - `0 = default` is a friction point with our international collaborators
+//!   who have moved away from this practice (notable exception: the UK met office)
```
pmlefeuvre-met, line 39:
```suggestion
//! - `0 = default` is a friction point with our international collaborators
//!   who nowadays aim to publish physical height instead, only Norway 
//!   and the UK Met. Office have not moved away from this practice.
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,168 @@
...
+//! if negative) of a sensor in centimetres.
+//!
+//! "Meteorological" in this case means that the height may not be the sensor's
+//! exact physical height, but a standard height close-by which has been
+//! approved for comparison to other stations. Also, for certain
+//! parameters, this height is not taken relative to the ground (i.e for wind
+//! it is taken relative to the top of any uneveness in the nearby landscape).
+//! Notably meteorological height differs from but is easily confused with:
+//! - elevation/height above sea level (hamsl) - typically refers to the vertical
+//!   distance between mean mean sea level and the Earth surface,
```
Lun4m, line 11:
```suggestion
//!   distance between mean sea level and the Earth surface,
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,168 @@
+//! Level as used here refers to meteorological height above ground (or depth
+//! if negative) of a sensor in centimetres.
+//!
+//! "Meteorological" in this case means that the height may not be the sensor's
+//! exact physical height, but a standard height close-by which has been
+//! approved for comparison to other stations. Also, for certain
+//! parameters, this height is not taken relative to the ground (i.e for wind
+//! it is taken relative to the top of any uneveness in the nearby landscape).
+//! Notably meteorological height differs from but is easily confused with:
+//! - elevation/height above sea level (hamsl) - typically refers to the vertical
```
Lun4m, line 10:
```suggestion
//! - elevation/height above mean sea level (hamsl) - typically refers to the vertical
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,168 @@
...
+//!   distance between mean mean sea level and the Earth surface,
+//!   in other words the ground the station stands on.
+//! - altitude - typically refers to the vertical distance between mean sea level
+//!   and the sensor at any point in the atmosphere (ex.: flight, radiosonde,...).
```
Lun4m, line 14:
```suggestion
//! - altitude - typically refers to the vertical distance between mean sea level
//!   and a sensor at any point in the atmosphere (ex.: flight, radiosonde, ...).
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,168 @@
...
+//!  
+//!
+//! This differs from previous level semantics at Met (notably in ODA, kvalobs,
+//! stinfosys), where level typically meant metres (though sometimes centimetres
+//! depending on the parameter and as defined by the value hlevel_scale).
+//! The level could not be 0 metre as 0 was used as a special value indicating that
```
Lun4m, line 20:
```suggestion
//! The level could not be 0 metre since 0 was used as a special value indicating that
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,168 @@
...
+//! the level is a default for the param (default value is defined in standard_hlevel
+//! as found in stinfosys), and NULL is always equivalent to 0 (when the data is
+//! meant to have a level, which not all data does).  In the previous semantics,
+//! negative integers were not allowed in kvalobs. Positive integers were given
+//! a direction from the stinfosys table sensorlevel defining:
```
Lun4m, line 25:
```suggestion
//! a direction in the stinfosys table sensorlevel which defines:
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,168 @@
...
+//! 1) height_above_ground in metre, 2) depth_below_surface in centimetre,
+//! and 3) depth_below_sea_surface in metre.
+//!
+//! Justifications for the old semantics (incomplete and inferred, as the
+//! people responsible are no longer with us):
```
Lun4m, line 30:
```suggestion
//! Justifications for the old semantics (incomplete and mostly inferred):
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,168 @@
...
+//! - `0 = default` reduces configuration work
+//! - `NULL = 0` means default levels can be omitted from messages reporting
+//!   reporting data from stations, saving some bytes, which used to be quite
```
Lun4m, line 33:
```suggestion
//! - `NULL = 0` means default levels can be omitted from messages
//!   reporting data from stations, saving some bytes, which used to be quite
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,168 @@
...
+//!   expensive.
+//!
+//! Justifications for changing to the new semantics:
+//! - `0 = default` encourages mistakes. A pattern observed at met is people
+//!   setting level = 0 for a new sensor to get it working/reporting quickly,
+//!   assuming someone will fix it later, which doesn't happen.
+//! - `0 = default` makes it impossible to represent a level that is genuinely
+//!   0, which does happen (i.e. surface parameters such as surface temperature)
```
Lun4m, line 41:
```suggestion
//! - `0 = default` makes it impossible to represent a level different than the default one that is genuinely
//!   0, which does happen (i.e. surface parameters such as surface temperature)
```

ingestion/src/levels.rs
```diff
@@ -0,0 +1,168 @@
...
+//! - `0 = default` is not what most end users want (although they got used to it)
+//!    meaning we either increase the burden on them, or convert to a different
+//!    scheme at request time, which is redundant and confusing.
+//! - `0 = default` is a friction point with our international collaborators
+//!   who nowadays aim to publish physical height instead, only Norway
+//!   and the UK Met. Office have not moved away from this practice.
```
Lun4m, line 47:
```suggestion
//! - `0 = default` is a friction point with our international collaborators
//!   who nowadays aim to publish physical height instead.
//!   Only Norway met and UK met have not moved away from this practice.
```

Lun4m (commented):
Found some typos and I propose some small changes

ingestion/src/levels.rs
```diff
@@ -0,0 +1,168 @@
...
+            &[],
+        )
+        .await?;
+
+    // build hashmap of param permits
+    let mut param_level = HashMap::new();
+
+    for row in rows {
+        let hlevel_scale: i32 = row.get(1);
+        if ![0, -2].contains(&hlevel_scale) {
```
intarga, line 125: @lcoram @Lun4m There is a bug here. The `!` should not be there. Might be a good idea to test this function before merging the next PR

intarga: fix in #113

PR #113, opened 2025-05-12
Kvkafka raw ingestion

intarga:


ingestion/src/kldata.rs
```diff
@@ -336,11 +332,7 @@ pub async fn filter_and_label_kldata<'a>(
                 }
             };
 
-            let (sensor, lvl) = in_datum
-                .id
-                .sensor_and_level
-                .map(|both| (Some(both.0), Some(both.1)))
-                .unwrap_or((None, None));
+            let (sensor, lvl): (i32, i32) = in_datum.id.sensor_and_level.unwrap_or((0, 0));
```
lcoram, line 336: So we are defaulting ourselves to 0,0, does this make the most sense or could it be handled differently? Are these always both not found? Couldn't I find a sensor and not have a level (or vice versa)?

ingestion/src/kldata.rs
```diff
@@ -484,23 +474,17 @@ mod tests {
...
-                )),
+                Err(ParseError::UnexpectedField("unexpected".to_string())),
                 "unexpected field",
             ),
             (
                 "kldata/messageid=10/type=501",
-                Err(Error::Parse(
-                    "missing field `nationalnr` in kldata header".to_string(),
-                )),
+                Err(ParseError::MissingField("nationalnr".to_string())),
```
lcoram, line 482: If don't have national number can we ingest as wigos or different id?

ingestion/src/legacy/checked.rs
```diff
@@ -0,0 +1,364 @@
...
+                        for level in sensor.levels {
+                            if let Some(kvdata) = level.kvdata {
+                                for kvdatum in kvdata {
+                                    data.push(UnlabelledDatum {
+                                        kvid: KvalobsId {
+                                            station: station.val,
+                                            paramid: kvdatum.paramid,
+                                            typeid: typeid.val,
+                                            sensor: sensor.val.unwrap_or(0),
+                                            level: level.val.unwrap_or(0),
```
lcoram, line 116: While it makes some sense to default to sensor 0 (its not without its problems also, since could end up with many timeseries all with sensor 0 which could cause confusion), could it make more sense to have a null level? should that be allowed?

ingestion/src/kldata.rs
lcoram: Ah ok!

ingestion/src/kldata.rs
lcoram: Ah, ok. Good that we ignore, it caused issues in ODA that stuff had overlapping stationids (but from different "namespaces").

intarga:
Getting a weird error in testing
```
2025-05-26T17:58:07.124729Z ERROR lard_ingestion::legacy::raw: Failed to insert kafka messages: postgres returned an error: db error: ERROR: deadlock detected
DETAIL: Process 153 waits for ShareLock on transaction 42479; blocked by process 154.
Process 154 waits for ShareLock on transaction 42482; blocked by process 153.
```

But I can't see anything that could cause a deadlock in the code, this error seems to be rare, and so difficult to reproduce :frowning_face:

intarga:
@Lun4m @lcoram 

Refined based on previous conversation:
- Fixed deadlock
- Subscribed to histkvalobs.checked
- Ingested non-float data from the raw queue to `public.nonscalar_data`

On a different note, I noticed a lot of empty values in obsinn messages on the raw queue, and I'm wondering if these are what Søren meant when he said that omitted columns should be deleted. This would be a lot easier to implement deletion for than what he originally described. I don't think we should tackle it in this PR, but I'll track it in an issue:
1. Check with Søren what this means
2. If it does indicate deletion, have a design meeting where we decide how to implement it

ingestion/src/legacy/checked.rs
```diff
@@ -0,0 +1,369 @@
...
+        let restricted_query = restricted_conn
+            .prepare(QUERY_STR)
+            .await
+            .expect("legacy::checked DB task couldn't prepare restricted query");
+
+        let mut raw_buffer: Vec<(Vec<UnlabelledDatum>, Offset)> =
+            Vec::with_capacity(DB_BUFFER_SIZE);
+
+        while db_rx.recv_many(&mut raw_buffer, DB_BUFFER_SIZE).await != 0 {
+            let offset = raw_buffer.last().unwrap().1.clone();
```
lcoram, line 274: Are we sure there is always a last? if it can be empty this is unsafe?

ingestion/src/kldata.rs
```diff
@@ -734,7 +719,7 @@ mod tests {
                                 param_code: "KLOBS".to_string(),
                                 sensor_and_level: None,
                             },
-                            value: NonScalar("20240910000000"),
+                            value: NonScalar("20240910000000".to_string()),
```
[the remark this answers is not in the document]
intarga: Not sure I understand what you mean? I'll try and answer anyway, apologies if I'm off the mark

KLOBS can't (or shouldn't anyway) be parsed as a float, since it represents a timestamp, so we store it in the db as a string (we have the table `public.nonscalar_data`) for this. We've considered making more tables for specific types (perhaps one with `DATETIMETZ` would be better for KLOBS), but we decided not to do this until we have a better overview of appropriate types for params (which has been low on the priority list)

There's no semantic change in this PR, just a change to types (the inner type of `ObsType::NonScalar` has been changed from `&str` to `String`, since in the kafka code we're sending it between threads it can't be a reference)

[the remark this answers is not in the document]
lcoram: What I am struggling with here is why a string that will always be 20240910000000 is being used and not a variable?

ingestion/src/legacy/checked.rs
intarga: Yes, from the [tokio docs](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv_many):

> For non-zero values of limit, this method will never return 0 unless the channel has been closed and there are no remaining messages in the channel’s queue. This indicates that no further values can ever be received from this Receiver. The channel is closed when all senders have been dropped, or when [close](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.close) is called.

Note that on the line above, we check `!= 0` to exit the loop in the case where the channel is closed

ingestion/src/kldata.rs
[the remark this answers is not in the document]
lcoram: Ah, yes I see... sorry I had not noted it was inside the test structure.

ingestion/src/legacy/raw.rs
```diff
@@ -0,0 +1,421 @@
...
+        res?;
+    }
+    drop(futures);
+
+    transaction.commit().await?;
+
+    Ok(())
+}
+
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
intarga: There is a case to be made for grouping some arguments into structs, like the prepared statements, or the metadata table. I don't think it's quite reached the point of being worth doing that yet though.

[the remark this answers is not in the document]
lcoram: Yes, I have had it complain to me about similar things.

ingestion/src/legacy/raw.rs
```diff
@@ -0,0 +1,421 @@
...
+                None => Param::Code(col.param_code),
+            };
+
+            let value: ObsType = if param_entry.is_some() && !param_entry.unwrap().is_scalar {
+                ObsType::NonScalar(val.to_string())
+            } else {
+                match val.parse::<f64>() {
+                    Ok(parsed) => ObsType::Scalar(parsed),
+                    Err(_) => ObsType::NonScalar(val.to_string()),
+                }
```
Lun4m, line 131: The parameter is supposed to be scalar, should the error branch actually return an error instead?
 
Edit: ah, we get to this else both if there is no `param_id` or if the parameter is a scalar, should they be treated the same?

ingestion/src/legacy/raw.rs
```diff
@@ -0,0 +1,421 @@
...
+
+            let (sensor, level) = col.sensor_and_level.unwrap_or((0, 0));
+
+            let val = val.trim();
+
+            // // TODO: is this right?
+            // // Ignore empty values
+            if val.is_empty() {
+                continue;
+            }
```
Lun4m, line 118: In the obsinn fixes branch I have something like [this](https://github.com/metno/lard/blob/9d92a6e5064db68c5ee01ff1707ec16344360d3f/ingestion/src/kldata.rs#L164-L171), maybe `Obstype` variants should be options?

ingestion/src/legacy/raw.rs
```diff
@@ -0,0 +1,421 @@
...
+    // to update the same two rows in a different order, although one part of
+    // this that doesn't make sense is that the raw query doesn't update, and so
+    // shouldn't acquire a row lock? The docs are unclear on this.
+    //
+    // Unfortunately, the introduction of this lock seems to cause a ~30%
+    // slowdown If we want to reclaim that throughput, alternative approaches
+    // might be:
+    // - Just let the deadlocks happen, then catch them and retry
+    // - Order the queries in each transaction in a consistent way (sort by
+    //   timeseries then obstime?) so such a deadlock can't happen
```
Lun4m, line 191: Would separating data in different vectors per obstype and using `COPY` have any benefit here?

ingestion/src/legacy/raw.rs
```diff
@@ -0,0 +1,421 @@
...
+                        error!("failed to poll raw kafka: {}", Error::Kafka(e));
+                    }
+                    Ok(message) => {
+                        metrics::counter!(KAFKA_RAW_MESSAGES_RECEIVED).increment(1);
+
+                        match message.payload() {
+                            Some(payload) => {
+                                if let Err(e) = 'parse_block: {
+                                    if !is_kldata_message(payload)? {
+                                        // TODO: explain
```
Lun4m, line 375: I guess this can be removed, or you can add `// Skipping BUFR messages`

ingestion/src/legacy/raw.rs
```diff
@@ -0,0 +1,421 @@
...
+                                        continue 'consume_loop;
+                                    }
+                                    // TODO: remove unwrap
```
Lun4m, line 378: Can be removed

ingestion/src/legacy/raw.rs
```diff
@@ -0,0 +1,421 @@
...
+                                            break 'parse_block Err(e.into())
+                                        },
+                                    }
+
+                                    Ok::<(), Error>(())
+                                } {
+                                    metrics::counter!(KAFKA_RAW_FAILURES).increment(1);
+                                    error!("failed to parse kldata message: {}", e);
+                                }
+                            },
```
Lun4m, line 397: We don't need the `if let(Err)`?

```suggestion
                            if !is_kldata_message(payload)? {
                                // TODO: explain
                                continue;
                            }
                            // TODO: remove unwrap
                            let payload_str = std::str::from_utf8(payload)?.trim();

                            let offset = Offset { partition:message.partition(), offset: message.offset() };

                            // TODO: remove clone?
                            match parse(payload_str, param_conversions.clone()){
                                Ok(data) => db_tx.send((data, offset)).await.unwrap(),
                                Err(e) => {
                                    error!("failed kldata message:{e}\n{:?}", payload_str);
                                    metrics::counter!(KAFKA_RAW_FAILURES).increment(1);
                                },
                            }
```

Lun4m (commented):
Some comments and maybe some changes?

ingestion/src/legacy/raw.rs
intarga: See noted intention to discuss with Søren. If these indicate deletion, we should actually do a deletion

ingestion/src/legacy/raw.rs
intarga: COPY is not option because it cannot handle conflicts

ingestion/src/legacy/raw.rs
Lun4m: I guess you'd have to expand everything that can error if you want to get rid of the extra indentation, just a matter of taste. Feel free to resolve

intarga: This version will kill the task if it encounters an error in parsing.

The point of the block and `if let` is to allow us to `?` all the errors without shortcutting the whole function (they only shortcut the block), so we can log/metric any errors, with only one call, instead of having to put 
```
if let Err(e) = ... {
    error!("error_message");
    metrics::counter!(KAFKA_RAW_FAILTURES).increment(1);
}
```
On everything that can error

ingestion/src/legacy/raw.rs
lcoram: I saw the note, and recall the discussion... I think they really are deletions, but obviously we need to clarify with søren.

Lun4m: Ah sorry!

ingestion/src/legacy/raw.rs
intarga: Yeah, IIRC when we originally wrote this parser we made plans to do a survey of all the params for this purpose, but it hasn't been a priority

ingestion/src/legacy/raw.rs
Lun4m: Right, we would need to know for sure that the raw queue inserts first

ingestion/src/legacy/raw.rs
Lun4m: Amazing! ⭐

Lun4m (approved):

PR #114, opened 2025-05-16
Update db schema and fix migrations import

Lun4m:
1. Remove the creation of duplicate timeseries for different dump files (dropping the `import_from` and `import_to` columns from `labels.kvalobs`)
1. Remove hash indexes
1. Convert level for the `labels.met` table when importing

### DIFF
--- a/db/001-public.sql
+++ b/db/001-public.sql
@@ -49,7 +49,6 @@ CREATE TABLE IF NOT EXISTS public.data (
     CONSTRAINT fk_data_timeseries FOREIGN KEY (timeseries) REFERENCES public.timeseries
 ) PARTITION BY RANGE (obstime);
 CREATE INDEX IF NOT EXISTS data_timestamp_index ON public.data (obstime);
-CREATE INDEX IF NOT EXISTS data_timeseries_index ON public.data USING HASH (timeseries);
 
 -- TODO: this should be renamed to 'public.text' or 'public.string'
 CREATE TABLE IF NOT EXISTS public.nonscalar_data (
@@ -61,4 +60,3 @@ CREATE TABLE IF NOT EXISTS public.nonscalar_data (
     CONSTRAINT fk_nonscalar_data_timeseries FOREIGN KEY (timeseries) REFERENCES public.timeseries
 ) PARTITION BY RANGE (obstime);
 CREATE INDEX IF NOT EXISTS nonscalar_data_timestamp_index ON public.nonscalar_data (obstime);
-CREATE INDEX IF NOT EXISTS nonscalar_data_timeseries_index ON public.nonscalar_data USING HASH (timeseries);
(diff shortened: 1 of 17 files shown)


migrations/kdvh/import/main.go
```diff
@@ -32,7 +32,10 @@ The following environement variables need to set:
 }
 
 func (config *Config) Execute() {
-	utils.GoMemLimitMessage("kdvh")
+	if os.Getenv("GOMEMLIMIT") == "" {
+		utils.PrintGoMemLimitMessage("kvalobs")
```
lcoram, line 36: Should this say kdvh?

migrations/lard/timeseries.go
```diff
@@ -14,7 +14,10 @@ type Label struct {
 	ParamID   int32
 	TypeID    int32
 	Sensor    *int32
-	Level     *int32
+	// Level converted to cm
```
lcoram, line 17: (and converting 0 default to an actual value)

migrations/utils/utils.go
```diff
@@ -11,13 +11,10 @@ import (
...
 
-func GoMemLimitMessage(packageName string) {
-	if os.Getenv("GOMEMLIMIT") == "" {
-		fmt.Println("To avoid OOM kills, set the GOMEMLIMIT environement variable.")
-		fmt.Println("For example:")
-		fmt.Printf("\tGOMEMLIMIT=4GiB ./migrate %s import ...\n", packageName)
-		os.Exit(1)
-	}
+func PrintGoMemLimitMessage(packageName string) {
+	fmt.Println("To avoid OOM kills, set the GOMEMLIMIT environement variable.")
```
lcoram, line 15: fix spelling? --> environement

lcoram (commented):
Some small changes, otherwise looks good

migrations/kdvh/import/main.go
Lun4m: Yes, done!

migrations/utils/utils.go
Lun4m: Done!

db/001-public.sql
```diff
@@ -49,7 +49,6 @@ CREATE TABLE IF NOT EXISTS public.data (
     CONSTRAINT fk_data_timeseries FOREIGN KEY (timeseries) REFERENCES public.timeseries
 ) PARTITION BY RANGE (obstime);
 CREATE INDEX IF NOT EXISTS data_timestamp_index ON public.data (obstime);
-CREATE INDEX IF NOT EXISTS data_timeseries_index ON public.data USING HASH (timeseries);
```
intarga, line 52: Since we know there's now an implicit index created by the unique constraint (which we want), I wonder if we should change the unique constraint to a primary key? It has the same effect, but I think it makes the intention behind the schema more obvious

intarga (approved):

db/001-public.sql
Lun4m: That's a good point, I changed all unique constraints to primary keys!

PR #115, opened 2025-05-23
Handle negative levels

lcoram:


ingestion/src/levels.rs
```diff
@@ -75,16 +75,18 @@ pub enum Error {
 
 #[derive(Debug, Clone)]
 pub struct Level {
-    hlevel: i32,
+    hlevel: Option<i32>,
     hlevel_scale: Option<i32>,
+    hlevel_type: Option<String>,
```
intarga, line 80: Why are these `Option`? And why is the new one `String`

Lun4m: Yeah, `hlevel_scale` should not have been an Option in the first place. The new one is String because it's TEXT in stinfosys?

lcoram: I see your point, we can 'wash' the data from stinfosys better as we take it in instead of handling it in the second function. 
Level was also originally not a function because we checked for it not being null in the SQL... We worried this might mean we didn't get scales/unit for stuff where there is no default.

intarga: > We worried this might mean we didn't get scales/unit for stuff where there is no default.

I think it's appropriate to replace `NULL` with `0`, right? It's consistent with the hlevel scheme's approach of `NULL` and `0` always being the same, and in this case I don't think it affects behaviour anyway

ingestion/src/levels.rs
```diff
@@ -109,10 +140,15 @@ pub async fn fetch_levels(stinfo_conn_string: &str) -> Result<ParamLevelTable, E
     });
 
     // query param table
+    // TODO: should we care about cases where standard_hlevel is NULL,
+    // while hlevel_scale is NOT NULL?
+    // Right now, for this case, we insert NULL level for every incoming level
```
Lun4m, line 145: We changed this to insert 0 instead of NULL if the incoming level is 0. Still unsure if it's the way to go 🤔

ingestion/src/levels.rs
```diff
@@ -145,24 +212,38 @@ pub fn param_get_level(
...
-    Ok(None)
+
+    // convert level to cm if currently in meters
+    if param_level.unit == Unit::M {
+        lvl *= 100
+    }
+
+    // convert to negative if 'Down'
+    if param_level.direction == Direction::Down {
+        lvl *= -1;
```
intarga, line 245: Maybe good to do `lvl.abs()` here? I know we absed the defaults from stinfosys, and the ones from kvalobs should be only positive, but it could be a good sanity check. Also because I'm unsure what convention obsinn follows.

ingestion/src/levels.rs
lcoram: Are we ok with this?

ingestion/src/levels.rs
lcoram: resolved?

ingestion/src/levels.rs
Lun4m: I guess? We can remove the todo I think

ingestion/src/levels.rs
```diff
@@ -145,24 +211,40 @@ pub fn param_get_level(
...
+    let Some(param_level) = level_table.get(&param_id) else {
+        warn!("could not find a scale for this param: {param_id}");
+        return Ok(None);
+    };
+
+    // If input level is already NULL, we simply insert NULL
+    // however, this should never happen since kvalobs / obsinn have default 0
+    let Some(mut lvl) = level else {
+        return Ok(None);
+    };
```
intarga, line 228: This is contingent on my change to the obsinn/checked parsers in the kafka raw PR that substitutes 0 where sensor level is missing. If you're going to remigrate based on this you should probably cherry-pick that commit

ingestion/src/levels.rs
```diff
@@ -145,24 +211,40 @@ pub fn param_get_level(
...
-    Ok(None)
+
+    // convert level to cm if currently in meters
+    if param_level.unit == Unit::M {
+        lvl *= 100
+    }
+
+    // convert to negative if 'Down'
+    if param_level.direction == Direction::Down {
+        // NOTE: should abs be done outside of this if, earlier?
```
intarga, line 244: I don't think it makes a difference

ingestion/src/levels.rs
intarga: https://github.com/metno/lard/pull/113/commits/6029b9335be65c0e3fe33d41d07820f8f0c3e5c9

Lun4m: I guess then we need `let mut lvl = level.unwrap_or(0)` here, or we could have it outside while making the function accept directly an `i32`

ingestion/src/levels.rs
```diff
@@ -121,20 +157,51 @@ pub async fn fetch_levels(stinfo_conn_string: &str) -> Result<ParamLevelTable, E
...
+        // convert the `sensorlevel_id` to a direction
+        // Down if `sensorlevel_id` from stinfosys contains the word "below"
+        let direction = match sensorlevel_id {
+            Some(s) => {
+                if s.contains("below") {
+                    Direction::Down
+                } else {
+                    Direction::Up
+                }
+            }
```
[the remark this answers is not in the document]
lcoram: Done

ingestion/src/levels.rs
lcoram: Done

ingestion/src/levels.rs
lcoram: Removed comment

ingestion/src/legacy/raw.rs
intarga, line 1: This pulled in a lot more than intended from my PR (I guess because of the file renaming), the commit is only a 24 line change, but this file is ingestion of the raw queue? This seems invalid anyway, because it has both the new files, and the old one they replace (kvkafka.rs).

We should drop this commit and either do a smaller cherry-pick, or just wait for the raw branch to be merged. Lmk if you want a hand with the git stuff, I'm not up to coming in to work again today, but we can pair it over video

ingestion/src/levels.rs
```diff
@@ -180,14 +177,16 @@ pub async fn fetch_levels(stinfo_conn_string: &str) -> Result<ParamLevelTable, E
...
 
         // convert the `sensorlevel_id` to a direction
-        // Down if `sensorlevel_id` from stinfosys contains the word "below"
+        // Down if `sensorlevel_id` from stinfosys is depth below surface or depth below sea surface
+        // Currently 3 values: "height_above_ground", "depth_below_surface", and "depth_below_sea_surface"
+        // need to change code if more are added to stinfosys
         let direction = match sensorlevel_id {
+            Some("height_above_ground") => Direction::Up,
+            Some("depth_below_surface") => Direction::Down,
+            Some("depth_below_sea_surface") => Direction::Down,
```
intarga, line 186: Can be written maybe more neatly as

```rust
Some("depth_below_surface") | Some("depth_below_sea_surface") => Direction::Down
```

Up to you which version you prefer though

ingestion/src/legacy/raw.rs
intarga: Sure, all that commit is anyway is a change to the function signature, and changes at the callsites to match

lcoram: I hope I have now fixed this...

ingestion/src/kvkafka.rs
```diff
@@ -287,7 +287,7 @@ async fn create_timeseries(
     let level = param_get_level(
         level_table.clone(),
         raw_datum.kvid.paramid,
-        raw_datum.kvid.level,
+        raw_datum.kvid.level.unwrap(),
```
intarga, line 290: I think this needs to be `unwrap_or(0)`, since we want to convert `None` to `0` instead of panicking when there's `None`. Also this just converts `None` level, but we should also convert `None` sensor to `0`.

Rather than converting it here, I think it's neater to change the definition of `KvalobsId` on lines 76 and 77 to remove the `Option`, and do the `.unwrap_or(0)` on lines 191 and 192 where we create the `KvalobsId`

intarga (approved):
:tada:

PR #118, opened 2025-05-26
Add IDF event endpoint

Lun4m:
I have left a lot of TODOs since I'm not 100% sure what we are supposed to do.
The main blockers for this to be merged are:
1. in ODA the filter timeseries is used to compute this report.
2. ~~We should move away from PT1M observations (which include approximate manual observations) to instead only use hourly automatic observations.~~

### DIFF
--- a/egress/src/patchwork.rs
+++ b/egress/src/patchwork.rs
@@ -46,17 +46,17 @@ pub struct Patch {
 }
 
 #[derive(Debug, Clone)]
-pub struct PatchworkTimeseriesTables {
+pub struct PatchworkTables {
     pub open: Arc<RwLock<PatchworkTimeseriesTable>>,
     pub restricted: Arc<RwLock<PatchworkTimeseriesTable>>,
 }
 
-impl PatchworkTimeseriesTables {
+impl PatchworkTables {
     pub fn new(
         open: PatchworkTimeseriesTable,
         restricted: PatchworkTimeseriesTable,
-    ) -> PatchworkTimeseriesTables {
-        PatchworkTimeseriesTables {
+    ) -> PatchworkTables {
+        PatchworkTables {
             open: Arc::new(RwLock::new(open)),
             restricted: Arc::new(RwLock::new(restricted)),
         }
@@ -69,7 +69,6 @@ pub struct MessagePriority {
     timerange: Timerange,
 }
 
-#[cfg(test)]
 impl MessagePriority {
     pub fn new(priority: i32, timerange: Timerange) -> MessagePriority {
         MessagePriority {
@@ -81,10 +80,10 @@ impl MessagePriority {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct MetLabel {
-    id: TsID,
+    id: i64,
     station_id: i32,
-    param_id: ParamID,
-    type_id: TypeID,
+    param_id: i32,
+    type_id: i32,
     level: Option<i32>,
     sensor: Option<i32>,
 }
@@ -92,10 +91,10 @@ pub struct MetLabel {
 #[cfg(test)]
 impl MetLabel {
     pub fn new(
-        id: TsID,
+        id: i64,
         station_id: i32,
-        param_id: ParamID,
-        type_id: TypeID,
+        param_id: i32,
+        type_id: i32,
         level: Option<i32>,
         sensor: Option<i32>,
     ) -> MetLabel {
@@ -114,7 +113,7 @@ impl MetLabel {
 // essentially removing the type_id from the label
 pub struct PatchworkLabel {
     pub station_id: i32,
-    pub param_id: ParamID,
+    pub param_id: i32,
     pub level: Option<i32>,
     // TODO: should this be optional??
     pub sensor: Option<i32>,
@@ -123,7 +122,7 @@ pub struct PatchworkLabel {
 impl PatchworkLabel {
     pub fn new(
         station_id: i32,
-        param_id: ParamID,
+        param_id: i32,
         level: Option<i32>,
         sensor: Option<i32>,
     ) -> PatchworkLabel {
@@ -139,17 +138,17 @@ impl PatchworkLabel {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct PriorityStruct {
     timerange: Timerange,
-    type_id: TypeID,
-    ts_id: TsID,
+    type_id: i32,
+    tsid: i64,
 }
 
 #[cfg(test)]
 impl PriorityStruct {
-    pub fn new(timerange: Timerange, type_id: TypeID, ts_id: TsID) -> PriorityStruct {
+    pub fn new(timerange: Timerange, type_id: i32, tsid: i64) -> PriorityStruct {
         PriorityStruct {
             timerange,
             type_id,
-            ts_id,
+            tsid,
         }
     }
 }
@@ -186,12 +185,7 @@ pub struct Fill {
 }
 
 impl Fill {
-    pub fn new(
-        from: DateTime<Utc>,
-        to: Option<DateTime<Utc>>,
-        tsid: TsID,
-        permit: PermitID,
-    ) -> Fill {
+    pub fn new(from: DateTime<Utc>, to: Option<DateTime<Utc>>, tsid: i64, permit: i32) -> Fill {
         Fill {
             from,
             to,
@@ -427,7 +421,9 @@ pub async fn fetch_patchwork_table(
     conn: &PooledPgConn<'_>,
     stinfosys_client: &Client,
 ) -> Result<PatchworkTimeseriesTable, Error> {
+    // TODO: this should be separate from the stinfosys stuff
     let db_ts_list = fetch_timeseries_list_from_database(conn).await?;
+
     let default_table = fetch_message_priority_default(stinfosys_client).await?;
     let exception_table = fetch_message_priority_exception(stinfosys_client).await?;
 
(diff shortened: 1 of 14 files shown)


egress/src/reports/idf_event.rs
```diff
@@ -0,0 +1,335 @@
...
+};
+
+use super::idf_station::IdfUnit;
+
+// sum(precipitation_amount PT1M)
+const PRECIPITATION_PARAM_ID: i32 = 105;
+
+// TODO: make sure these defaults are correct
+const DEFAULT_LEVEL: Option<i32> = Some(200);
+const DEFAULT_SENSOR: Option<i32> = Some(0);
```
Lun4m, line 25: @ketilt Can you double check that these are the parameters used in ODA to calculate the IDF event? I don't remember seeing them explicitly referenced in the old code

egress/src/patchwork.rs
```diff
@@ -30,10 +30,16 @@ pub type MessagePriorityDefaultTable = HashMap<(TypeID, ParamID), MessagePriorit
...
 // type for list of applicable timeseries
-pub type ApplicableTimeseriesList = Vec<(i64, DateTime<Utc>, DateTime<Utc>)>;
+pub type ApplicableTimeseriesList = Vec<Patch>;
 /// This table contains the patchworked timeseries, mapping to typeid and timeseriesid
 pub type PatchworkTimeseriesTable = HashMap<PatchworkLabel, Vec<Fill>>;
 
+pub struct Patch {
+    pub tsid: TsID,
+    pub from: DateTime<Utc>,
+    pub to: DateTime<Utc>,
```
lcoram, line 40: Can to be open in these patches? aka an option...

egress/src/patchwork.rs
```diff
@@ -591,13 +614,20 @@ pub fn get_applicable_timeseries(
         });
         // have overlap
         if let Some(t) = overlap {
-            applicable_ts.push((f.tsid, t.from.unwrap(), t.to.unwrap_or(to)));
+            applicable_ts.push(Patch {
+                tsid: f.tsid,
+                from: t.from.unwrap(),
+                to: t.to.unwrap_or(to),
```
lcoram, line 620: I guess we use the users to date, so its not open or we have to use the current time if they didn't specify?

egress/src/reports/idf_event.rs
lcoram: 200 is default level for air temp params, but 1000 would be default for wind stuff... (off the top of my head). Sensor 0 is definetely the default (most recomended).

egress/src/reports/idf_event.rs
```diff
@@ -0,0 +1,335 @@
...
+    // TODO: are these timeseries ordered by fromtime?
+    let mut futures = patches
+        .iter()
+        .map(|patch| async move {
+            conn.query(
+                "SELECT obstime, corrected \
+                FROM legacy.data \
+                WHERE timeseries = $1 \
+                    AND corrected IS NOT NULL \
+                    AND quality_code != 7 \
```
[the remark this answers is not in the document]
lcoram: Only 7 was considered rubbish?

lcoram (approved):
Maybe an e2e test for the endpoint as mentioned by @intarga 
Also, might be good to sanity check the actual mechanics of the algorithm @ketilt

egress/src/reports/idf_event.rs
intarga: Level is kind of irrelevant for precip though, right?

lcoram: Good point... yeah, I don't think its relevant

egress/src/reports/idf_event.rs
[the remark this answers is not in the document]
Lun4m: @ketilt probably knows best?

egress/src/reports/idf_event.rs
intarga: @pmlefeuvre-met  is the authority on this, but IIRC precipitation does not have a significant relationship with level. Precipitation at ground level should be functionally indistinguishable from, say, 10m up

egress/src/patchwork.rs
Lun4m: Right now I think you are required to send from/to times in the request, so to can't be open

egress/src/reports/idf_event.rs
```diff
@@ -0,0 +1,378 @@
...
+    label.param_id == PRECIPITATION_PARAM_ID
+        && label.level == DEFAULT_LEVEL
+        && label.sensor == DEFAULT_SENSOR
+}
+
+pub async fn idf_event_availability_handler(
+    State(patchwork_tables): State<PatchworkTables>,
+) -> Result<Json<IdfEventAvailability>, (StatusCode, String)> {
+    // TODO: need to implement this also for restricted?
+    let ot = patchwork_tables.open.read().map_err(internal_error)?;
```
lcoram, line 267: Need to pass roles here to also deliver restricted? 
Will have same weakness as patchwork avaiable, in that if you are allowed to see any restricted ts you will see all of them listed here (but you cannot get the data). 
We can try to filter by roles here, should also implement same filtering in patchwork available.

integration_tests/tests/end_to_end.rs
```diff
@@ -285,6 +283,7 @@ async fn test_latest_endpoint() {
 
 #[tokio::test]
 async fn test_patchwork_endpoint_failure() {
+    // FIXME: this is running against an empty patchwork table
```
lcoram, line 286: has this been fixed?

util/src/deserialize.rs
```diff
@@ -0,0 +1,62 @@
+//! Here we define various functions that can be used with `#[serde(deserialize_with = "")]`
+use std::{fmt, marker::PhantomData, str::FromStr};
+
+use serde::{
+    de::{self, Visitor},
+    Deserialize, Deserializer,
+};
+
+// TODO: could use serde_with but it seems a bit annoying since it introduces new macros/syntax
```
lcoram, line 9: Can this be used in patchwork, to simplify the code?

egress/src/patchwork.rs
```diff
@@ -572,42 +585,65 @@ pub fn create_patchwork_timeseries_table(
...
+        let patches = get_applicable_timeseries(from, to, label, tables.restricted)?;
+        if !patches.is_empty() {
+            return Ok((patches, pools.restricted));
+        }
+    };
+
+    let patches = get_applicable_timeseries(from, to, label, tables.open)?;
+    Ok((patches, pools.open))
+}
+
```
intarga, line 607: I don't think this function should exist, we decided to get rid of it in the patchwork PR. Control flow should be pushed up the call stack, and returning pools from functions seems like a bad idea.

egress/src/patchwork.rs
```diff
@@ -76,21 +80,21 @@ impl MessagePriority {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct MetLabel {
-    id: TsID,
+    id: i64,
     station_id: i32,
-    param_id: ParamID,
-    type_id: TypeID,
+    param_id: i32,
+    type_id: i32,
```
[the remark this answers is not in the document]
lcoram: Funny, I just discussed this with @Lun4m. He was thinking that they are most needed when it's not clear what the field is (like in some of the type definitions with tuples), but not in struct since then it's already clear. I argued that say, particularly for tsid it's good to be explicit so no one makes a mistake and tries to use i32... As well as if we ever have to change it, you change the type and not all the other spots.

intarga (changes requested):

egress/src/patchwork.rs
Lun4m: Yeah, I'm not sure I can return a connection without leaking it. Is this better? Directly in the idf event handler
```rust
    let (patches, conn) = 'label: {
        if roles.is_some() {
            let patches = patchwork::get_applicable_timeseries(
                params.fromtime,
                params.totime,
                idf_event_label,
                tables.restricted,
            )
            .map_err(internal_error)?;

            if !patches.is_empty() {
                let conn = pools.restricted.get().await.map_err(internal_error)?;
                break 'label (patches, conn);
            }
        };

        let patches = patchwork::get_applicable_timeseries(
            params.fromtime,
            params.totime,
            idf_event_label,
            tables.open,
        )
        .map_err(internal_error)?;
        let conn = pools.open.get().await.map_err(internal_error)?;

        (patches, conn)
    };
```
I will have to copy paste it in the windorose handler too

integration_tests/tests/common/mod.rs
```diff
@@ -243,7 +268,29 @@ pub async fn create_db_pools() -> DbPools {
...
+pub fn empty_patchwork_tables() -> PatchworkTables {
+    PatchworkTables::new(HashMap::new(), HashMap::new())
+}
+
+pub async fn update_patchwork_table(
+    conn: &PooledPgConn<'_>,
+    table: Arc<RwLock<PatchworkTimeseriesTable>>,
+) {
+    let db_list = fetch_timeseries_list_from_database(conn).await.unwrap();
+    let message_prioity = mock_message_priority();
```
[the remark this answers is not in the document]
Lun4m: Done

egress/src/patchwork.rs
Lun4m: That looks way better, yeah, thanks! And returning an `Option` for data as you say it's probably going to make it even cleaner

egress/src/reports/idf_event.rs
Lun4m: Done

egress/src/patchwork.rs
Lun4m: Unfortunately "chain" methods don't work with async (maybe on nightly?), but you are right, I think we can `unwrap_or_default` the roles and we won't hit the database if we have the wrong permits

integration_tests/tests/end_to_end.rs
Lun4m: Fixed now

intarga (approved):

PR #129, opened 2025-07-17
try to fix ansible lint

lcoram:


Lun4m:
I think we should fix the ansible-lint version in `.github/workflows/ci.yml` to the same one we have in `ansible/requirements.txt` instead.
```yaml
      - name: Run ansible-lint
        uses: ansible/ansible-lint
        with:
          gh_action_ref: v24.9.2
          setup_python: "true"
          working_directory: "ansible"
          requirements_file: "requirements.yml"
```
It seems a bit silly that all the variables inside roles need to be prefixed even if they are "private". There's an open issue in the ansible lint repo regarding this change.

Lun4m (approved):

PR #152, opened 2025-09-22
Skip timeseries with no permit

Lun4m:
egress currently crashes if a timeseries does not have any permit


Lun4m:
Yep, will fix those and merge! :ok:

PR #153, opened 2025-09-26
only unauthorized if have a bearer token that is invalid

lcoram:


egress/src/auth.rs
```diff
@@ -110,17 +110,13 @@ pub async fn auth_middleware(
...
-        "malformed authorization header".to_string(),
-    ))?;
-
-    // if errors, the token is invalid
-    let roles = verify_token(&token, certs).map_err(error::unauthorized)?;
+    // check if bearer format
+    if let Some(token) = parse_auth_header(auth_header) {
+        // if errors, the token is invalid
+        let roles = verify_token(&token, certs).map_err(error::unauthorized)?;
+        req.extensions_mut().insert(Some(roles));
```
lcoram, line 117: Do we need to insert None if we don't hit this part of the code...?

intarga:
A blurb providing context for this change would be appreciated

lcoram:
If this gets merged and deployed hopefully the endpoint that gets patchwork in lard.frost-dev might start working... Currently it errors (if I do basic auth, could test with bearer)

intarga:
> only unsure if that is the last problem we will find with Frost.

This should be able to be tested in staging, or better yet locally. If that seems inconvenient, then it's a signal to work on improving our tooling and infrastructure until it isn't

lcoram:
I have tested that if I use oauth2 (aka keyword bearer in the auth header) then frost works, whereas it gets the error I solved in this request if I use basic. Will try to bring up both lard and frost locally to test more fully.

lcoram:
In order to test this locally I think I need to write a bigger docker compose setup with the egress, and potentially include a frost image so we could test that (although I could run that seperately if I expose the egress on localhost)... Does this make sense, or am I going down the wrong path?

PR #167, opened 2025-10-16
Obsinn message parsing fixes

intarga:
Supersedes #65 

### TODO
- [x] make sure ingestion errors are logged

### DIFF
--- a/ingestion/src/kldata.rs
+++ b/ingestion/src/kldata.rs
@@ -39,20 +39,14 @@ pub enum ParseError {
 }
 
 /// FIXME: these params are scalar in Stinfosys, but are not when coming from Obsinn.
-pub const SPECIAL_CASES: [&str; 7] = [
+pub const SPECIAL_CASES: [&str; 6] = [
     // METAR params that come in as 'xxL' and 'xxR', where 'x' is a numeric character.
     // We need to decide how to treat them (Kvalobs silently discards them apparently)
     // Or if they need to be changed in Stinfosys
-    "X1R",
-    "X2R",
-    "X3R",
-    "WS",
-    "WS2",
+    "X1R", "X2R", "X3R", "WS", "WS2",
     // seems to be a number most of the times, but gets an 'a' every once in a while.
     // Maybe it's in hex format?
     "W1",
-    // with value "N70.18664E30.05277", can maybe be separated into lat and lon params?
-    "STN#META##LatitudeLongitude",
     // Leaving the following disabled, since even though I've seen them with these weird
     // outputs, they do seem to primarily send float data. These weird values seem to be
     // their elementids?
--- a/ingestion/src/legacy/raw.rs
+++ b/ingestion/src/legacy/raw.rs
@@ -117,13 +117,13 @@ fn parse_obs(
                 None => Param::Code(col.param_code.clone()),
             };
 
-            let value: ObsType = if (param_entry.is_some() && !param_entry.unwrap().is_scalar)
+            let value: ObsType = if param_entry.is_some() && param_entry.unwrap().is_scalar
                 // things marked as scalar in stinfosys that are known not to be floats
-                || kldata::SPECIAL_CASES.contains(&col.param_code.as_str())
+                && !kldata::SPECIAL_CASES.contains(&col.param_code.as_str())
             {
-                parse_nonscalar(val)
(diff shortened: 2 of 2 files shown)


ingestion/src/kldata.rs
```diff
@@ -38,6 +38,29 @@ pub enum ParseError {
...
+    // Maybe it's in hex format?
+    "W1",
+    // with value "N70.18664E30.05277", can maybe be separated into lat and lon params?
+    "STN#META##LatitudeLongitude",
+    // with value "Flatetemperatur"
+    "TSS",
+    // with value "windspeed10m"
+    "FF_01",
+    // with value "winddir10m"
+    "DD_01",
```
Lun4m: These are not floats?

ingestion/src/kldata.rs
```diff
@@ -153,6 +176,39 @@ pub fn parse_columns(cols_raw: &str) -> Result<Vec<ObsinnId>, ParseError> {
...
+                    parsed as float. Consider adding it to SPECIAL_CASES",
+                col.param_code, val
+            );
+            ObsType::NonScalar(Some(val.to_string()))
+        }
+    }
+}
+
+pub fn parse_nonscalar(val: &str) -> ObsType {
+    if val.is_empty() || val == "-" {
```
lcoram: Do we need the /// here too?

ingestion/src/lib.rs
```diff
@@ -148,14 +148,14 @@ impl FromRef<IngestorState> for Arc<HashMap<(i32, RelativeDuration), rove::Pipel
...
-    NonScalar(String),
+    Scalar(Option<f64>),
+    NonScalar(Option<String>),
 }
 
 pub struct Datum {
     timeseries_id: i64,
     // needed for QC
-    param_id: i32,
+    param_id: Option<i32>,
```
lcoram: Wait, no the param_id can still be none?

lcoram (approved):
Looks like general improvements, and if the trimming tests pass it seems like its good to get this in.

ingestion/src/kldata.rs
```diff
@@ -38,6 +38,29 @@ pub enum ParseError {
...
+    "X1R",
+    "X2R",
+    "X3R",
+    "WS",
+    "WS2",
+    // seems to be a number most of the times, but gets an 'a' every once in a while.
+    // Maybe it's in hex format?
+    "W1",
+    // with value "N70.18664E30.05277", can maybe be separated into lat and lon params?
+    "STN#META##LatitudeLongitude",
```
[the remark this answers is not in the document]
intarga: Are you saying we should filter out all of them? I think it makes senses to try and ingest for now, if we later find out there's no use we can always delete them. This one does seem useful though

ingestion/src/kldata.rs
intarga: I posted the values they were sent with above, I at least have no idea how to parse those as floats.

ingestion/src/kldata.rs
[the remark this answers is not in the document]
intarga: I mean, I don't see a huge need to differentiate the two, ultimately we're just telling the parser to treat them as nonscalars

ingestion/src/legacy/raw.rs
```diff
@@ -110,26 +112,18 @@ fn parse_obs(
...
+                None => Param::Code(col.param_code.clone()),
             };
 
-            let value: ObsType = if param_entry.is_some() && !param_entry.unwrap().is_scalar {
-                ObsType::NonScalar(val.to_string())
+            let value: ObsType = if (param_entry.is_some() && !param_entry.unwrap().is_scalar)
+                // things marked as scalar in stinfosys that are known not to be floats
+                || kldata::SPECIAL_CASES.contains(&col.param_code.as_str())
+            {
+                parse_nonscalar(val)
```
Lun4m: I think this if condition should be reversed, it tries to parse a scalar even when there's no param, giving rise to wrong logs.

```suggestion
            let value: ObsType = if param_entry.is_some() && param_entry.unwrap().is_scalar
                // things marked as scalar in stinfosys that are known not to be floats
                && !kldata::SPECIAL_CASES.contains(&col.param_code.as_str())
            {
                parse_scalar(val)
```

PR #169, opened 2025-10-16
Null paramid bug

lcoram:
Found one spot we appear to create labels which could have null paramid. Unclear of ramifications of not creating this label, they would still have their more "native" label and still get a timeseriesid.

### DIFF
--- a/ingestion/src/legacy/common.rs
+++ b/ingestion/src/legacy/common.rs
@@ -7,7 +7,6 @@ use chrono::{DateTime, Utc};
 use futures::{stream::FuturesOrdered, StreamExt};
 use serde::Deserialize;
 use thiserror::Error;
-use tracing::warn;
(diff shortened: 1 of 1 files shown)


ingestion/src/legacy/common.rs
```diff
@@ -152,22 +153,29 @@ async fn create_timeseries<T: Clone>(
...
+                    &level, // currently just overrriding the level in the met label
+                    &raw_datum.kvid.sensor,
+                ],
+            )
+            .await?;
+    } else {
+        warn!(
+            "Not creating labels.met for \"{:?}\" since paramid does not exist",
+            raw_datum.kvid.param
+        );
```
intarga: This warn will be too noisy to put in production IMO

intarga (approved):

PR #179, opened 2025-10-31
avoid bug where fail when trying to parse nulls

lcoram:
```
Oct 31 11:24:46 lard-b1 lard_egress[225014]: thread 'tokio-runtime-worker' panicked at /home/louiseo/Documents/gitlab.met.no/lard/egress/src/patchwork.rs:706:28:
Oct 31 11:24:46 lard-b1 lard_egress[225014]: error retrieving column 2: error deserializing column 2: a Postgres value was `NULL`
```
triggered with this call: http://xxxx:3000/patchwork?stationids=17810&paramids=329&levels=200&sensors=0&from=2025-09-01T00:00:00Z&to=2025-10-31T00:00:00Z

Seems pluviometer data is sending nulls...

### DIFF
--- a/egress/src/patchwork.rs
+++ b/egress/src/patchwork.rs
@@ -146,8 +146,8 @@ impl PriorityStruct {
 pub struct PatchworkDatum {
     // can assume have original and timestamp? (the field for original value
     // can technically be null (database schema wise)...)
-    // but do not always have corrected and quality code
-    original: f64,
+    // definitely do not always have corrected and quality code (eg. before qc'ing)
+    original: Option<f64>,
     timestamp: DateTime<Utc>,
     corrected: Option<f64>,
     quality_code: Option<i32>,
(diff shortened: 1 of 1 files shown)


intarga:
~~2 design considerations:~~

~~1. Corrected can also be NULL, should that be checked too?~~

~~2. Would it make more sense to wrap original and corrected in `Option`?~~

Nvm, I looked and saw that corrected is already an `Option`, but would it make sense for original to also be `Option`?

lcoram:
Well I would say it depends on what we see patchwork as...
**Patchwork's primary purpose it to remove typeid, and deliver one timeseries for that simplified label.** 
If that is its sole / primary purpose, then passing out exactly what is in the database (as in wrapping all potential nulls as options) is the better solution. 
However, if we see patchwork as anything more than that... as in, something that should avoid giving users things that are in the database that will likely make no sense to them (such as these null values), then the first solution with the change to the SQL is better.

